PR #4, opened 2015-01-18
Adding support for datadog monitors

bluelion:
Added monitor support, note no schedules supported yet


zorkian:
Oh, nice! Thank you for doing this. (And sorry for the delay -- was travelling.)

bluelion:
Most welcome. I'll be adding the timeboard code some time soon as well since I'll be using the API for that too.

PR #6, opened 2015-05-18
Ensure body is closed after reading

seiffert:
Not closing a HTTP response's body leads to a memory leak


zorkian:
Nice catch!

seiffert:
thanks for merging (and for maintaining this library!)

PR #9, opened 2015-06-26
Metric.Type in JSON must be key `metric_type`

parkr:
According to the API documentation as of today, this is the proper key.

http://docs.datadoghq.com/api/#metrics-post

/cc @zorkian

### DIFF
--- a/series.go
+++ b/series.go
@@ -17,7 +17,7 @@ type DataPoint [2]float64
 type Metric struct {
 	Metric string      `json:"metric"`
 	Points []DataPoint `json:"points"`
-	Type   string      `json:"type"`
+	Type   string      `json:"metric_type"`
 	Host   string      `json:"host"`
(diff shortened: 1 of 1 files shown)


parkr:
That being said, this official library contradicts the docs: https://github.com/DataDog/dogapi-rb/blob/master/lib/dogapi/v1/metric.rb#L70

zorkian:
Did you test it with this change, does it work as expected? They might have renamed it, but still allow both, and haven't updated all libraries yet.

parkr:
@zorkian This is based on the API documentation I linked to above. Perhaps @degemer can shed some light here?

degemer:
After some digging, `type` is the correct option, we're going to update the docs shortly. Thanks for the warning!

parkr:
Thanks, @degemer !!

zorkian:
Thanks for the update everybody!

PR #11, opened 2015-08-06
Fix API endpoint for inviting new users

seiffert:
`api/v1/account/invite` results in HTTP status code 404. 

According to [the docs](http://docs.datadoghq.com/api/#users), the correct endpoint is `api/v1/invite_users`.


zorkian:
Thanks for the fix -- and the doc link to make it easy to review ;)

PR #14, opened 2015-08-20
Build against 1.4, 1.5 and tip

ojongerius:
But still allow failures building tip.


ojongerius:
Thanks for the quick merge! :+1:

Issue #17, opened 2015-10-09
How do you query metrics?


winhowes:
In datadog's api they mention you can query metrics (http://docs.datadoghq.com/api/#metrics-query). Looking through your docs, I couldn't find anything to do that. I saw `SearchMetrics`, but that didn't query timeseries data.

zorkian:
Looks like nobody has implemented that method yet. Should be pretty trivial though, about 5 lines of code in `series.go` if you want to take a crack at it.

winhowes:
I've made a PR, tested it against my own stuff and found it works: https://github.com/zorkian/go-datadog-api/pull/18

zorkian:
Merged -- thank you! Closing this :)

PR #18, opened 2015-10-09
Add query metrics method

winhowes:
I've added a query metrics method to implement http://docs.datadoghq.com/api/#metrics-query. I have tested it out on some of my own projects and it successfully fetches back the data. There's also some minor clean up here based on go patterns. This addresses #17

### DIFF
--- a/series.go
+++ b/series.go
@@ -8,6 +8,8 @@
 
 package datadog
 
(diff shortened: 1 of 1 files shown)


zorkian:
Thank you so much! Awesome :)

PR #19, opened 2015-10-22
Add CRUD user management functions

jvehent:
This is a new feature of the Datadog API that is not yet document but available in their Python client. It allows to retrieve, update and delete users. Creation must still happen via invitation.

Sample code:
```go
package main

import (
        "fmt"
        "log"
        "os"

        "github.com/zorkian/go-datadog-api"
)

func main() {
        client := datadog.NewClient(os.Getenv("DATADOG_API_KEY"), os.Getenv("DATADOG_APP_KEY"))

        user, err := client.GetUser("bob@example.net")
        if err != nil {
                log.Fatal(err)
        }
        fmt.Printf("%+v\n", user)

        user.Name = "Bob Kelso"
        user.Disabled = false
        user.IsAdmin = true
        err = client.UpdateUser(user)
        if err != nil {
                log.Fatal(err)
        }

        user, err = client.GetUser(user.Handle)
        if err != nil {
                log.Fatal(err)
        }
        fmt.Printf("%+v\n", user)

        user.Disabled = true
        user.IsAdmin = false
        err = client.UpdateUser(user)
        if err != nil {
                log.Fatal(err)
        }
        user, err = client.GetUser(user.Handle)
        if err != nil {
                log.Fatal(err)
        }
        fmt.Printf("%+v\n", user)

}

### DIFF
--- a/users.go
+++ b/users.go
@@ -8,6 +8,16 @@
 
 package datadog
 
+type User struct {
+	Handle   string `json:"handle"`
+	Email    string `json:"email"`
+	Name     string `json:"name"`
+	Role     string `json:"role"`
+	IsAdmin  bool   `json:"is_admin"`
+	Verified bool   `json:"verified"`
+	Disabled bool   `json:"disabled"`
+}
+
 // reqInviteUsers contains email addresses to send invitations to.
 type reqInviteUsers struct {
 	Emails []string `json:"emails"`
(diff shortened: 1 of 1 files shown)


zorkian:
Thanks for the patch -- sorry for the slow merge!

PR #20, opened 2015-11-05
Use DATADOG_HOST for the URL if ENV variable defined.

darron:
Just wondering if you'd consider merging this.

I know it has very limited application, but at Datadog - where I work - we have more than 1 possible target and I'd love to use this properly in one of our alternate environments.

I'm pretty new to Go - so if you'd like it done a different way please let me know.

Thanks for a great library - has helped me out with some internal tooling I'm hoping to open source if it works well.

### DIFF
--- a/request.go
+++ b/request.go
@@ -15,17 +15,24 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"strings"
 )
 
 // uriForAPI is to be called with something like "/v1/events" and it will give
 // the proper request URI to be posted to.
 func (self *Client) uriForAPI(api string) string {
+	var url = ""
+	if os.Getenv("DATADOG_HOST") != "" {
+		url = os.Getenv("DATADOG_HOST")
+	} else {
+		url = "https://app.datadoghq.com"
+	}
 	if strings.Index(api, "?") > -1 {
-		return "https://app.datadoghq.com/api" + api + "&api_key=" +
+		return url + "/api" + api + "&api_key=" +
 			self.apiKey + "&application_key=" + self.appKey
 	} else {
-		return "https://app.datadoghq.com/api" + api + "?api_key=" +
+		return url + "/api" + api + "?api_key=" +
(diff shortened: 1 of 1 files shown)


zorkian:
You work at Datadog and you're submitting a pull request to my library? That is super exciting! I love Datadog. :D

Anyway, I absolutely would consider this. Do you need it to be in an environment variable, though? Else, I would suggest adding a `apiHost` string to the `Client` structure and then creating a `NewCustomClient` function where you can specify the endpoint to talk to and it sets it.

It seems like it might be cleaner that way rather than expecting everybody to use the same environment variable approach, which is probably very specific to a given infrastructure.

Thoughts?

darron:
The reason I did it that way, is because most of our other libraries already support environment variables to switch targets:

https://github.com/DataDog/datadogpy/blob/master/datadog/__init__.py#L73
https://github.com/DataDog/dogapi-rb/blob/master/lib/dogapi/common.rb#L153

I can take a look at doing it the way you're asking - will give it a shot.

zorkian:
Okay -- no that's fine. Consistency is good, and if the other libraries do it this way, we can certainly do the same. I'll review & merge!

request.go
```diff
@@ -15,17 +15,24 @@ import (
...
 
 // uriForAPI is to be called with something like "/v1/events" and it will give
 // the proper request URI to be posted to.
 func (self *Client) uriForAPI(api string) string {
+	var url = ""
+	if os.Getenv("DATADOG_HOST") != "" {
+		url = os.Getenv("DATADOG_HOST")
+	} else {
+		url = "https://app.datadoghq.com"
+	}
```
zorkian: For simplicity:

```
url := os.Getenv("DATADOG_HOST")
if url == "" {
    url = "https://app.datadoghq.com"
}
```

darron:
Good call - that's nicer and cleaner for sure.

Thanks!

zorkian:
Thanks for the fix! :)

darron:
No problem at all - thanks for the library!

PR #21, opened 2015-11-06
Return body on error.

ojongerius:
Which will result in:

```Got error:  API error 400 Bad Request: {"errors":["The value provided for parameter 'query' is invalid"]}``` 

vs the old

```Got error:  API error 400```

### DIFF
--- a/request.go
+++ b/request.go
@@ -12,6 +12,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: request.go [hunk 1]
```diff
@@ -60,7 +61,11 @@ func (self *Client) doJsonRequest(method, api string,
 	defer resp.Body.Close()
 
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		return errors.New("API error: " + resp.Status)
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+		return errors.New(fmt.Sprintf("API error %s: %s", resp.Status, body))
```
zorkian: Instead please use: https://golang.org/pkg/fmt/#Errorf

`return fmt.Errorf("API error %s: %s", resp.Status, body))`

zorkian:
Thank you! :)

PR #22, opened 2015-11-06
Fix datadog upstream changes

ojongerius:
Very inelegant workaround of Datadog API changes. Passing a metric alert with a struct containing an empty Monitor struct (which will have treshold values default to 0) now throws a "400 Bad Request: {"errors":["Critical threshold (0) does not match that used in the query (0)."]}). This broke my provider.

Note sure if you want this in, I just needed a quick fix for terraform-provider-datadog. Using embeds would be nicer, but I could not find a quick way to get that in without breaking backwards compatibility.

### DIFF
--- a/monitors.go
+++ b/monitors.go
@@ -41,6 +41,28 @@ type Monitor struct {
 	Options Options `json:"options"`
 }
 
+// Metric Monitor cannot have default zero values for thresholds
+//notifying your team when some defined threshold is exceeded.
+type MetricMonitor struct {
+	Id      int     `json:"id"`
+	Type    string  `json:"type"`
+	Query   string  `json:"query"`
+	Name    string  `json:"name"`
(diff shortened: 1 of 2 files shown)


zorkian:
So the problem is just that if we send a 0 value the API errors? If so, we should just add `omitempty` to the JSON descriptors. That way the created JSON object does not include the 0 value fields with that flag. Can you give that a shot and tell me if it works?

If not, I'll ping Datadog and ask them to fix their API. That error message does seem bogus.

ojongerius:
@zorkian thanks for the feedback it was too late to think clear and remember "omitempty": :+1: 

You can debate about the legitimacy of the behaviour. Having one endpoint for many different monitors, where some options are meaningful surely makes using it a little awful.

zorkian:
Yeah, for sure. Thanks for the update!

PR #23, opened 2015-12-21
Fix typo introduced in refactor.

ojongerius:
This was introduced by my own doings, and took a little while to show up as it is silently dropped by the Datadog API. Sorry about that.


zorkian:
Thanks for catching it :)

PR #24, opened 2015-12-21
Prevent errors unmarshalling returned monitor object containting sile…

ojongerius:
Fixes 2 things:

* Using decimals for thresholds is supported, but using it breaks unmarshalling.
* "Silenced" maps use strings for keys, and ints as values.

Note this change will break clients expecting these values to be integers. The upshot is that we can now use monitors with decimal thresholds, which is a perfectly good use good.

### DIFF
--- a/monitors.go
+++ b/monitors.go
@@ -9,25 +9,26 @@
 package datadog
 
 import (
+	"encoding/json"
 	"fmt"
 )
 
 type ThresholdCount struct {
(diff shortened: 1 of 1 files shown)


zorkian:
I'm fine with breaking clients, since this is a compiled language and they can just upgrade. It won't silently break someone's system. (Unless they have something crazy, I suppose.)

ojongerius:
:bow:

PR #29, opened 2015-12-28
Add markers to timeseries widgets

seiffert:
Markers are constants drawn as horizontal lines in a widget. Sadly, this feature isn't documented in the official API docs. However, we're using it since a couple of months successfully.


zorkian:
No official documentation? Huh, alright.

PR #30, opened 2015-12-28
Retry failed HTTP requests to Datadog API with exponential backoff

seiffert:
We experience a lot of failed requests to the Datadog API that are answered with 5xx and (false) 405 HTTP responses. To circumvent this issue, we'd like this client library to retry API calls in such cases.

Due to the non-idempotent nature of POST requests, we can't retry them in a sane matter. Therefore, they are excluded from this mechanism.
For the retrying, we use a library called `backoff` that implements exponential backup up until a maximum elapsed time (which is the time spent for all tries). For now, I'm hardcoding the maximum elapsed time to 60 seconds.

### DIFF
--- a/request.go
+++ b/request.go
@@ -11,12 +11,16 @@ package datadog
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
+	"time"
+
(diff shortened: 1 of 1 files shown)


zorkian:
In theory aren't many of the POST methods idempotent? I.e. if you're submitting a set of metrics, resubmitting them should just work. Also with things like "delete X"...

It'd probably be pretty hard and confusing to try to get that all right, though.

PR #31, opened 2015-12-28
Allow accessing a time series aggregator

seiffert:
This attribute is returned when querying for metrics as [described in the docs](http://docs.datadoghq.com/api/#metrics-query). Its value is something like `"min"` or `"max"`.


zorkian:
Thank you for the pull requests -- appreciated!

PR #33, opened 2016-01-12
Add support for downtime api

rfairburn:
This adds support and tests for the /v1/downtime api as outlined at http://docs.datadoghq.com/api/.


zorkian:
Thanks for doing -- and with the test!

PR #38, opened 2016-03-01
Add support for checks API

yayalice:
http://docs.datadoghq.com/api/#service_checks

Example usage:
```go
client := datadog.NewClient(apiKey, appKey)
client.PostCheck(datadog.Check{
    Check:    "check name",
    Status:   datadog.CRITICAL,
    Message:  "check message",
    Tags:     []string{"tag1:val1", "tag2:val2"},
    HostName: "hostname",
})
```

### DIFF
--- a/checks.go
+++ b/checks.go
@@ -1,0 +1,25 @@
+package datadog
+
+type Check struct {
+	Check     string   `json:"check"`
+	HostName  string   `json:"host_name"`
+	Status    status   `json:"status"`
+	Timestamp string   `json:"timestamp,omitempty"`
+	Message   string   `json:"message,omitempty"`
+	Tags      []string `json:"tags,omitempty"`
+}
+
+type status int
+
+const (
+	OK status = iota
+	WARNING
+	CRITICAL
+	UNKNOWN
+)
+
+// PostCheck posts the result of a check run to the server
+func (client *Client) PostCheck(check Check) error {
+	return client.doJsonRequest("POST", "/v1/check_run",
+		check, nil)
+}
--- a/checks_test.go
+++ b/checks_test.go
@@ -1,0 +1,22 @@
+package datadog_test
+
+import (
+	"testing"
+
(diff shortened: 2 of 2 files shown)


checks.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+type status int
+
+const OK status = 0
+const WARNING status = 1
+const CRITICAL status = 2
+const UNKNOWN status = 3
+
+// PostMetrics takes as input a slice of metrics and then posts them up to the
```
zorkian: This comment seems a bit incorrect?

zorkian:
Looks great, just fix the comment and I can merge it? Thank you!

checks.go
```diff
@@ -0,0 +1,24 @@
...
+	HostName  string   `json:"host_name"`
+	Status    status   `json:"status"`
+	Timestamp string   `json:"timestamp,omitempty"`
+	Message   string   `json:"message,omitempty"`
+	Tags      []string `json:"tags,omitempty"`
+}
+
+type status int
+
+const OK status = 0
```
seiffert: A more idiomatic way of representing this enum-like set of constants would be something like 

```go
const (
	OK status = iota
	WARNING
	CRITICAL
	UNKNOWN
)
```

see https://golang.org/doc/effective_go.html#constants for more information.

zorkian: I agree that's idiomatic, but in this case OK must always be exactly 0, WARNING must be 1, etc. Does `iota` guarantee it will assign counting from 0?

yayalice: @seiffert good suggestion
> [the value of iota] is reset to 0 whenever the reserved word const appears in the source

(from https://golang.org/ref/spec#Iota)

checks.go
yayalice: oops copy / paste.

zorkian:
Thank you for doing this! :)

PR #40, opened 2016-03-22
add Creator field to Monitor struct

shawnps:
Hi @zorkian,

I'm trying to access the creator of a monitor, so I added this struct field. Let me know if there are any style changes you'd like me to make.

### DIFF
--- a/monitors.go
+++ b/monitors.go
@@ -34,6 +34,7 @@ type Options struct {
 //Monitors allow you to watch a metric or check that you care about,
 //notifying your team when some defined threshold is exceeded.
 type Monitor struct {
+	Creator creator  `json:"creator"`
 	Id      int      `json:"id,omitempty"`
 	Type    string   `json:"type,omitempty"`
 	Query   string   `json:"query,omitempty"`
@@ -41,6 +42,14 @@ type Monitor struct {
 	Message string   `json:"message,omitempty"`
 	Tags    []string `json:"tags,omitempty"`
 	Options Options  `json:"options,omitempty"`
+}
+
+// creator is the creator of the monitor
+type creator struct {
+	Email  string
+	Handle string
+	ID     int
+	Name   string
 }
 
 // reqMonitors receives a slice of all monitors
--- a/monitors.go
+++ b/monitors.go
@@ -48,7 +48,7 @@ type Monitor struct {
 type Creator struct {
 	Email  string `json:"email,omitempty"`
 	Handle string `json:"handle,omitempty"`
-	ID     int    `json:"id,omitempty"`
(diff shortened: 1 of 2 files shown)


monitors.go
```diff
@@ -43,6 +44,14 @@ type Monitor struct {
 	Options Options  `json:"options,omitempty"`
 }
 
+// creator is the creator of the monitor
+type creator struct {
```
seiffert: Can you maybe expose the `creator` struct by writing it with a capital `C`? Otherwise, it's not possible to construct `Monitor`s with `Creator`s outside of the `datadog` package.

monitors.go
```diff
@@ -43,6 +44,14 @@ type Monitor struct {
...
+	Email  string
```
seiffert: please add tags with the JSON attribute names to the properties of `creator`

monitors.go
```diff
@@ -34,6 +34,7 @@ type Options struct {
 //Monitors allow you to watch a metric or check that you care about,
 //notifying your team when some defined threshold is exceeded.
 type Monitor struct {
+	Creator creator  `json:"creator"`
```
seiffert: the `creator` attribute should probably be also `omitempty` in JSON

shawnps:
@seiffert please take another look when you have a chance, I think I got all of your suggestions in but let me know if there's anything else

seiffert:
Thanks, Looks good to me!

monitors.go
```diff
@@ -43,6 +44,14 @@ type Monitor struct {
 	Options Options  `json:"options,omitempty"`
 }
 
+// Creator is the creator of the monitor
+type Creator struct {
+	Email  string `json:"email,omitempty"`
+	Handle string `json:"handle,omitempty"`
+	ID     int    `json:"id,omitempty"`
```
zorkian: Could this be `Id` to match the other uses (such as in Monitor)?

zorkian:
Sorry to nitpick the capitalization, but it'll bug me otherwise... :)

While you're in here, can you update `.travis.yml` and remove the line that says `  - 1.4`? We want to stop testing against Go 1.4 since `golint` now requires 1.5+ and that test will never pass.

monitors.go
shawnps: Done

shawnps:
> Sorry to nitpick the capitalization, but it'll bug me otherwise... :)

No problem :)



> While you're in here, can you update .travis.yml and remove the line that says - 1.4? We want to stop testing against Go 1.4 since golint now requires 1.5+ and that test will never pass.

Done

zorkian:
Great, thank you so much both of you!

PR #41, opened 2016-03-28
Support CreateUser API

### LINKED ISSUE #37: CreateUser API
Hi guys,

The [API reference for DD](http://docs.datadoghq.com/api/) shows a `POST api/v1/user` to create a new user. This client has the `invite_users` call in its stead.

I assume the DD API changed over time and that's why it's not here. Any reason for me not to add it in with a PR?

Thanks.

seiffert:
Datadog introduced an API to create users that differs from the (undocumented) one we currently support (https://github.com/zorkian/go-datadog-api/blob/master/users.go#L28). This PR adds a method `CreateUser` that allows users to add user accounts based on an email address and a name.

fixes #37

## Todo: 

- [x] Write integration tests
- [ ] Find a way to automatically clean up test users

### DIFF
--- a/integration/users_test.go
+++ b/integration/users_test.go
@@ -1,0 +1,42 @@
+package integration
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func init() {
+	client = initTest()
+}
+
+func TestCreateAndDeleteUser(t *testing.T) {
+	handle := "test@example.com"
+	name := "tester"
+
+	user, err := client.CreateUser(handle, name)
+	if err != nil {
+		t.Fatalf("Failed to create user: %s", err)
+	}
+	if user == nil {
+		t.Fatalf("CreateUser did not return an user.")
+	}
+
+	defer func() {
+		err := client.DeleteUser(handle)
+		if err != nil {
+			t.Fatalf("Failed to delete user: %s", err)
+		}
+	}()
+
+	assert.Equal(t, user.Handle, handle)
+	assert.Equal(t, user.Name, name)
+
+	newUser, err := client.GetUser(handle)
+	if err != nil {
+		t.Fatalf("Failed to get user: %s", err)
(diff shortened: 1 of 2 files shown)


seiffert:
I've added a commit with a basic integration test that executes `CreateUser`, `GetUser`, `DeleteUser`. While running this test, I noticed, that the `DeleteUser` API call does not actually delete the user but disables it. Therefore, I wonder

1. whether `DeleteUser` is maybe named incorrectly and 
2. how to test user creation and deletion in an automated way. The only way to actually delete the user I found was to cancel the users invitation (which is not a documented API call...). 

Any thoughts on that, @zorkian?

seiffert:
Also, I recommend deprecating the `InviteUsers` method as it uses an undocumented API.

zorkian:
So, it seems DataDog doesn't want to support deleting users: https://help.datadoghq.com/hc/en-us/articles/203765615-How-do-I-setup-or-remove-a-team-member-in-Datadog-

When you run the integration test do you use a test account or are you running it against your main/prod account (and hoping it cleans up appropriately)?

seiffert:
Oh, thanks for the link, I wasn't aware of that... 
I'm using a second account for testing.

zorkian:
What's the status here? Do you want this merged, even though it just disables? (Does the create/disable flow work multiple times in a row?)

zorkian:
Ping?

ojongerius:
Ping!

seiffert:
Sorry, we are in the progress of migrating our metrics somewhere else which is why I'm no longer working on the Datadog integration.
Do you want to take over from here or should I close this PR?

zorkian:
I'll just merge it then as it's got a test and adds functionality. Thank you for all your contributions over time!

ojongerius:
@seiffert just out of curiosity, what are you migrating to?

PR #43, opened 2016-04-06
Docstring should match method name.

ojongerius:


zorkian:
Thank you for the doc fixes!

Issue #49, opened 2016-06-03
Failed to unmarshal json because 'no_data_timeframe' sometimes returns false


takus:
Hi,

I hit the error `json: cannot unmarshal bool into Go value of type int.` when I called GetMonitors(). It seems that the value of `options.NoDataTimeframe` becomes `false`.

According to the official document, `no_data_timeframe` supposed to be `int`.

> no_data_timeframe the number of minutes before a monitor will notify when data stops reporting. Must be at least 2x the monitor timeframe for metric alerts or 2 minutes for service checks.

But, it becomes sometimes `bool`. For example, `GET ALL MONITOR DETAILS` of Shell section in http://docs.datadoghq.com/api/ describes it like this:

```
    {
        "id": 91875,
        "message": "",
        "name": "**system.net.bytes_rcvd** over **host:host0** was **> 100** on average during the **last 1h**.",
        "options": {
            "escalation_message": "",
            "no_data_timeframe": false,
            "notify_audit": true,
            "notify_no_data": false,
            "renotify_interval": null,
            "silenced": {},
            "timeout_h": null
        },
        "org_id": 1499,
        "query": "avg(last_1h):sum:system.net.bytes_rcvd{host:host0} > 100",
        "type": "metric alert",
        "multi": false,
        "created": "2015-12-18T16:34:14.014039+00:00",
        "modified": "2015-12-18T16:34:14.014039+00:00"
    }
```

I'm asking Datadog supports whether this is expected or not. If this is not a bug, I'll make a pull request to manage this issue.

zorkian:
Thanks for the report! Did you get any answer on the issue yet?
 
--
Mark Smith
mark@qq.is
 
 
On Thu, Jun 2, 2016, at 05:25 PM, Takumi Sakamoto wrote:
> Hi,
> I hit the error json: cannot unmarshal bool into Go value of type int.
> when I called GetMonitors(). It seems that the value of
> options.NoDataTimeframe becomes 'false'.
> According to the official document, no_data_timeframe should be int.
>> no_data_timeframe the number of minutes before a monitor will notify
>> when data stops reporting. Must be at least 2x the monitor timeframe
>> for metric alerts or 2 minutes for service checks.
> But, it becomes sometimes false. For example, GET ALL MONITOR
> DETAILS of Shell section in http://docs.datadoghq.com/api/ describes
> it like this:


>     { "id": 91875, "message": "", "name": "**system.net.bytes_rcvd**
[...]

PR #50, opened 2016-06-17
Define NoDataTimeframe type for unmarshaling JSON correctly

takus:
Patch for #49.
Replace value into 0 if 'false' value is found.


zorkian:
It looks like the `gofmt` test is failing? I can't divine why by eyeballing the diff, but can you fix up the format and repush?

takus:
Sorry, I fixed up format.

zorkian:
Thanks for the fix!

PR #55, opened 2016-08-21
Fix UnmarshalJSON for NoDataTimeframe

yfronto:
* Unmarshal function was checking if the type was a string and then assigning zero but never assigned the proper value
* Added parse to demarshal properly


zorkian:
Nice catch, thank you!

PR #58, opened 2016-10-03
Refactor the request options

paroxp:
In order to make some changes to the struct and not break the tests, we’d like to separate single Request into its own struct, which can be later referenced anywhere.

We've added two of some missing options in the Request struct, followed by the integration test case implementation, for this very occasion.

We've also created a [pull request](https://github.com/hashicorp/terraform/pull/9178), to the Terraform library, which would need to be merged roughly in the same time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ mostly used for automating dashboards/alerting and retrieving data (events, etc)
 
 The source API documentation is here: <http://docs.datadoghq.com/api/>
 
+
 ## USAGE
 
 To use this project, include it in your code like:
@@ -24,7 +25,7 @@ Then, you can work with it:
 
 ``` go
     client := datadog.NewClient("api key", "application key")
-    
+
     dash, err := client.GetDashboard(10880)
     if err != nil {
         log.Fatalf("fatal: %s\n", err)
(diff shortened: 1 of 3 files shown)


paroxp:
Hello!

We're in rather awkward position to be honest.

The Terraform tests are failing, because of this Pull Request not being merged. 
Merging straight away, would fix the tests, *HOWEVER* would brake the terraform master branch.

We may need to have a common understanding with you and `hashicorp`, as to when we can all merge in the same time, to avoid problems.

Thanks.

zorkian:
That's fine with me. I'm on Freenode as "zorkian" if someone wants to ping me and let me know a good time to do it. (I'm travelling and in the GMT+8 timezone for the next two weeks, but should be reasonably reachable if I have a point of contact on IRC.)

paroxp:
Hello again.

We've just discovered, the terraform has this library in their vendor directory "cached". Therefore, if we go ahead with this merge, it shall not break the master and will in fact fix the tests.

However, now it appears that travis is failing on `go1.5` as the command below, is unable to install the dependency.

```
go get -u github.com/golang/lint/golint
```

Not sure, if something we'd need to act on, as the tests are passing.

stack72:
Hey all

I work on the TF team. Luckily, we vendor our deps (as said above). If you ping me when you need the PR merged, I can run the tests and pull the trigger :)

P.

paroxp:
Hello @stack72!

Thanks for joining the conversation. Because, you vendor the deps, it may be better to have this PR merged first, by @zorkian.

As soon as it is merged, we will update the branch/pr in the terraform repo, with the new version of this repo.

Thanks again guys,
Rafal.

stack72:
👍

zorkian:
Thanks for the heads up! This has now been merged.

PR #51, opened 2016-08-11
Dashboard: Add Aggregator and Type to Graph struct.

ojongerius:

PR #59, opened 2016-10-22
Improve timeboard support

nyanshak:
* Add better support for different graph types, such as supporting yaxis
  and markers on timeseries graphs, as well as precision and custom
  units on query type graphs.

### DIFF
--- a/dashboards.go
+++ b/dashboards.go
@@ -22,6 +22,23 @@ type GraphDefinitionRequest struct {
 	Style              struct {
 		Palette string `json:"palette,omitempty"`
 	} `json:"style,omitempty"`
+
+	// For change type graphs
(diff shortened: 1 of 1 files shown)


nyanshak:
One of the builds fails because golint is not installed:
```
The command "go get -u github.com/golang/lint/golint" exited with 2.
0.01s$ golint ./...
/home/travis/build.sh: line 45: golint: command not found
The command "golint ./..." exited with 127.
0.03s$ test `gofmt -l . | wc -l` = 0
```

This is similar to Pull request #51. I am trying to add more complete support for all the graph types, to hopefully be used in the terraform datadog provider later.

zorkian:
Thanks for the patch! Yeah, I need to clean up the builds. golint decided to stop supporting go 1.5 and we're still testing against it.

PR #60, opened 2016-10-25
Add complete event struct for timeboards

nyanshak:
* The event struct didn't have any fields, and now it does, so it should
  now be possible to read / write events in timeboards.

PR #61, opened 2016-10-25
Move event struct into timeboard graph definition

nyanshak:
Fixes issue with PR #60 where event struct was in base of the graph struct instead of graph.Definition.


zorkian:
What's better than a PR? A fix for the PR very shortly after :) Thanks for being quick!

Issue #56, opened 2016-08-25
Consider updating library to use pointers for optional fields


ojongerius:
Hi there,

So far using default values has worked for us but it will not work in the case where values can either be set or not, and not setting them can influence behaviour.

An example [issue](https://github.com/hashicorp/terraform/issues/8055) issue I ran into  `require_full_window` which is a Boolean, and optional. If the option is not set Datadog uses custom behaviour, which is desirable in some cases.

Using the current API this field defaults to false, which is _different_ to it not being set. I suspect, but have not checked if there are other cases where we ran into this issue.

A solution is to refactor the library to use pointers, as this allows us to check if a value is actually set. The upside is more precision, the downside is that clients using the library now have to do a little more work to use pointers. 

A struct like `Monitor` currently looks like so:

``` golang
// Monitor allows watching a metric or check that you care about,
// notifying your team when some defined threshold is exceeded
type Monitor struct {
    Creator Creator  `json:"creator,omitempty"`
    Id      int      `json:"id,omitempty"`
    Type    string   `json:"type,omitempty"`
    Query   string   `json:"query,omitempty"`
    Name    string   `json:"name,omitempty"`
    Message string   `json:"message,omitempty"`
    Tags    []string `json:"tags,omitempty"`
    Options Options  `json:"options,omitempty"`
}
```

And would look like this using pointers for optional fields only:

``` golang
// Monitor allows watching a metric or check that you care about,
// notifying your team when some defined threshold is exceeded
type Monitor struct {
    Creator *Creator  `json:"creator,omitempty"`
    Id      *int      `json:"id,omitempty"`
    Type    string   `json:"type,omitempty"`
    Query   string   `json:"query,omitempty"`
    Name    *string   `json:"name,omitempty"`
    Message *string   `json:"message,omitempty"`
    Tags    []string `json:"tags,omitempty"`
    Options *Options  `json:"options,omitempty"`
}
```

Or like so using pointers for all fields:

``` golang
// Monitor allows watching a metric or check that you care about,
// notifying your team when some defined threshold is exceeded
type Monitor struct {
    Creator *Creator  `json:"creator,omitempty"`
    Id      *int      `json:"id,omitempty"`
    Type    *string   `json:"type,omitempty"`
    Query   *string   `json:"query,omitempty"`
    Name    *string   `json:"name,omitempty"`
    Message *string   `json:"message,omitempty"`
    Tags    []string `json:"tags,omitempty"`
[...]

PR #62, opened 2016-11-03
Fixes for API inconsistency and pointers for some optional structs

nyanshak:
* Adds a complete event struct for timeboards
* Moves list of event structs into Graph.Definition (instead of in Graph directly)
* Change some instances of float64 to json.Number since in some cases Datadog API returns a number and in other cases a string.
* Change some struct attributes to pointers for optional fields

### DIFF
--- a/dashboards.go
+++ b/dashboards.go
@@ -55,9 +55,9 @@ type Graph struct {
 
 		// For timeseries type graphs
 		Yaxis struct {
-			Min   float64 `json:"min,omitempty"`
-			Max   float64 `json:"max,omitempty"`
-			Scale string  `json:"scale,omitempty"`
+			Min   *float64 `json:"min,omitempty"`
+			Max   *float64 `json:"max,omitempty"`
+			Scale *string  `json:"scale,omitempty"`
 		} `json:"yaxis,omitempty"`
 
 		// For query value type graphs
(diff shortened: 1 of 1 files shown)


nyanshak:
Only test failing is the linter on Go 1.5.

I'd suggest doing one of these to stop running into these build failures:
- remove Go 1.5 from list of versions to build in travis
- Change .travis.yml to allow failures of go 1.5:
```
matrix:
    allow_failures:
        - go: tip
        - go: 1.5
```
- Something like this in .travis.yml:
```
script:
  - if [[ "${TRAVIS_GO_VERSION}" != "1.5" ]]; then go get -u github.com/golang/lint/golint ; fi
  - if [[ "${TRAVIS_GO_VERSION}" != "1.5" ]]; then golint ./... ; fi
  - test `gofmt -l . | wc -l` = 0
  - make test
```

nyanshak:
@zorkian ping :)

zorkian:
Sorry for the slow response here. My understanding of this diff as well as your comment on #56 is that these fields are just optional fields, and in that case, using pointers for just the optionals seems a reasonable compromise.

I'm still curious for feedback on the code snippet I posted on that issue if you have some time to look and comment, though.

PR #63, opened 2016-11-14
Correcting the event fields for adding an event

caledhwa:
The fields on "event" for pushing types and source are either missing or incorrectly named. The reference on correct naming was taken from here:

https://github.com/DataDog/dd-agent/blob/7946ff9d8f1fa4276a99d38cb53df2e67118efad/checks.d/kubernetes.py

I am trying to impersonate a type and found that setting these values correctly (in my case `kubernetes`) produced the desired results.

### DIFF
--- a/events.go
+++ b/events.go
@@ -25,10 +25,11 @@ type Event struct {
 	AlertType   string   `json:"alert_type,omitempty"`
 	Host        string   `json:"host,omitempty"`
 	Aggregation string   `json:"aggregation_key,omitempty"`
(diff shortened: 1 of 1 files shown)


zorkian:
Thanks for tracking these down and fixing them!

caledhwa:
@zorkian - Thanks for such a quick turn around on this! I'll be off deleting my fork for now. If I find anything else like this, I'll submit another PR. Thanks again!

PR #64, opened 2016-11-14
Adding methods to get by name and tag.

caledhwa:
In order to find a monitor to prevent duplicating it, I needed some way to perform query searches by Name and/or Tag. These methods basically recreate the query you can do in the UI for monitors.

I tried to run the test suite locally, but it seemed to hang - so I did not add integration tests for these methods. However, if you need me to, I can try and get those integration tests running correctly and add additional tests for `GetMonitorsByName` and `GetMonitorsByTag`

Let me know if you need updates and I'll jump on them. I'd like to get this in the library ASAP - if you have time. :)

### DIFF
--- a/monitors.go
+++ b/monitors.go
@@ -11,7 +11,9 @@ package datadog
 import (
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strconv"
+	"strings"
 )
 
 type ThresholdCount struct {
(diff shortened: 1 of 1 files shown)


zorkian:
Nope this is fine -- integration tests are great but I don't require them. I'll see if I can figure out why it was hanging.

PR #65, opened 2016-11-24
make checks.go/status public as its exposed in the API

GlenDC:
Renamed the type alias `status` to `Status` to make it public.

> Why?

Because it's exposed within the `Check` struct, yet it was private and thus people wishing to provide a function that takes a `status` value, couldn't actually do that as the type itself was private. This is pretty bad practice. When a value is exposed, its type should be exposed as well.


zorkian:
Thanks for the patch -- sorry for the slow merge, holidays here!

PR #68, opened 2016-12-06
Rename receiver remove redundant lines

ojongerius:
* Refactor receiver names to have some relationship to what the represent
* If error is the only value returned, we might as well test for the value on the same line.
* Add testing for 1.7, remove 1.5


zorkian:
This is also better idiomatic Go rather than Self. Thanks!

PR #77, opened 2017-01-17
Allow configuring a timeout

cwood:
Right now we default to 60 seconds this should be configurable. So we can set this to a bigger/smaller number.


ojongerius:
LGTM 👍

zorkian:
Perfect, thank you!

PR #78, opened 2017-01-18
Add rate limit support

cwood:
This will return back a error if the rate limiting is down to 0. On some APIs datadog will rate limit and it doesnt make sense to keep trying. This will also return it back as a error. Im not sure if this is good or bad. I feel like it should come back as `data, ratelimit, err := client.QueryMetrics...`

Currently the way this will work is on either doRequest or doRequestWithRetries it will check for the header if it has it and returns back a error it will then check the remaining if it is at 0 it will then return the rate limit error as a error type.

The above should be fine and give the user some information. I should probably also update the doc strings for the 4ish endpoints that I know have rate limiting via the datadog api.

### DIFF
--- a/request.go
+++ b/request.go
@@ -17,10 +17,18 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/cenkalti/backoff"
+)
+
+const (
+	RateLimitHeader     = "X-RateLimit-Limit"
+	RatePeriodHeader    = "X-RateLimit-Period"
+	RateRemainingHeader = "X-RateLimit-Remaining"
+	RateResetHeader     = "X-RateLimit-Reset"
 )
 
 // uriForAPI is to be called with something like "/v1/events" and it will give
@@ -73,6 +81,12 @@ func (client *Client) doJsonRequest(method, api string,
 	}
 	defer resp.Body.Close()
 
+	if err := client.getRateLimit(resp); err != nil {
+		if err.(*RateLimit).Remaining == 0 {
+			return err
+		}
+	}
+
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
 		body, err := ioutil.ReadAll(resp.Body)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


request.go
```diff
@@ -73,6 +81,12 @@ func (client *Client) doJsonRequest(method, api string,
 	}
 	defer resp.Body.Close()
 
+	if err := client.getRateLimit(resp); err != nil {
+		if err.(*RateLimit).Remaining == 0 {
```
ojongerius: Would be 🏆 if we could back start backing off, if we have _X-RateLimit-Reset_ we know how long we'd have to wait for, right?

request.go
```diff
@@ -120,6 +155,12 @@ func (client *Client) doRequestWithRetries(req *http.Request, maxTime time.Durat
 			return err
 		}
 
+		if err := client.getRateLimit(resp); err != nil {
+			if err.(*RateLimit).Remaining == 0 {
```
ojongerius: See other note about backing off instead of giving up.

ojongerius:
> This will return back a error if the rate limiting is down to 0. On some APIs datadog will rate limit and it doesnt make sense to keep trying.

Can we not wait until the next period instead of giving up? See my inline comment.

>This will also return it back as a error. Im not sure if this is good or bad. I feel like it should come back as data, ratelimit, err := client.QueryMetrics...

I think your implementation is fine if it uses the X-RateLimit headers to calculate how to back off a little smarter.

request.go
cwood: yea I wasnt sure about that. it seems like something the author should be aware of then do there own retry vs the library doing it for them. Mostly since this can be a large amount of time since once you are out you could wait as long as a hour before you can retry again.

request.go
```diff
@@ -104,6 +118,27 @@ func (client *Client) doJsonRequest(method, api string,
 	return nil
 }
 
+func (client *Client) getRateLimit(response *http.Response) error {
+
+	ratelimit := response.Header.Get(RateLimitHeader)
+	if ratelimit == "" {
```
cwood: It doesn't supper matter the header.I just wanna verify one of them is there and then create the rate limit object to return back.

request.go
```diff
@@ -104,6 +118,27 @@ func (client *Client) doJsonRequest(method, api string,
...
+		return nil
+	}
+
+	rateperiod := response.Header.Get(RatePeriodHeader)
+	rateremaining := response.Header.Get(RateRemainingHeader)
+	ratereset := response.Header.Get(RateResetHeader)
+
+	limit, _ := strconv.Atoi(ratelimit)
+	period, _ := strconv.Atoi(rateperiod)
```
cwood: Could do that. I'll leave it up the to author to decide.

request.go
cwood: Yea I dont think we should do that. But I am up for discussion.

ojongerius:
Great to have some of this code in here, and we can start improving on this if we see fit. Thanks @cwood!

cwood:
haha I didnt know you were a writer to this. I can work on that stuff tomorrow AM once I get in local SF time.

ojongerius:
🏆

PR #81, opened 2017-01-25
Fix dashboard acceptance tests

jszwedko:
Weren't up-to-date with the defined API resulting in the following errors:

```
# command-line-arguments
integration/dashboards_test.go:167: cannot use "warm" (type string) as type *string in assignment
integration/dashboards_test.go:192: cannot use 99.9 (type float64) as type json.Number in field value
integration/dashboards_test.go:196: cannot use 99 (type int) as type json.Number in field value
integration/dashboards_test.go:200: cannot use 99 (type int) as type json.Number in field value
FAIL    command-line-arguments [build failed]
```

As a result of #62 which changed the `datadog.GraphDefinitionRequest` to use a pointer for `Style`, it seemed prudent to break this out into a separate struct to make it easier for people to initialize (just fixing the above compilation errors ended up causing a null pointer panic in `TestCreateAndDeleteAdvancesTimeseriesDashboard`.

Issue #73, opened 2016-12-07
Monitor integration test fails


ojongerius:
`TestCreateAndDeleteMonitor` currently fails:

```
=== RUN   TestCreateAndDeleteMonitor
--- FAIL: TestCreateAndDeleteMonitor (69.00s)
        Error Trace:    monitors_test.go:21
        Error:          Not equal: &datadog.Monitor{Creator:datadog.Creator{Email:"", Handle:"", Id:0, Name:""}, Id:1352361, Type:"metric alert", Query:"avg(last_15m):avg:system.disk.in_use{*} by {host,device}
                                != &datadog.Monitor{Creator:datadog.Creator{Email:"ojongerius@atlassian.com", Handle:"ojongerius@atlassian.com", Id:155928, Name:""}, Id:1352361, Type:"metric alert", Query:"avg

                        Diff:
                        --- Expected
                        +++ Actual
                        @@ -1,6 +1,6 @@
                        -(*datadog.Monitor)(0xc4206617a0)({
                        +(*datadog.Monitor)(0xc420661c20)({
                          Creator: (datadog.Creator) {
                        -  Email: (string) "",
                        -  Handle: (string) "",
                        -  Id: (int) 0,
                        +  Email: (string) (len=24) "ojongerius@atlassian.com",
                        +  Handle: (string) (len=24) "ojongerius@atlassian.com",
                        +  Id: (int) 155928,
                           Name: (string) ""
                        @@ -12,3 +12,4 @@
                          Message: (string) (len=12) "Test message",
                        - Tags: ([]string) <nil>,
                        + Tags: ([]string) {
                        + },
                          Options: (datadog.Options) {

        Error Trace:    monitors_test.go:27
        Error:          Not equal: &datadog.Monitor{Creator:datadog.Creator{Email:"", Handle:"", Id:0, Name:""}, Id:1352361, Type:"metric alert", Query:"avg(last_15m):avg:system.disk.in_use{*} by {host,device}
                                != &datadog.Monitor{Creator:datadog.Creator{Email:"ojongerius@atlassian.com", Handle:"ojongerius@atlassian.com", Id:155928, Name:""}, Id:1352361, Type:"metric alert", Query:"avg

                        Diff:
                        --- Expected
                        +++ Actual
                        @@ -1,6 +1,6 @@
                        -(*datadog.Monitor)(0xc4206617a0)({
                        +(*datadog.Monitor)(0xc42063c900)({
[...]

PR #82, opened 2017-01-25
Add support for new_host_delay in monitor definitions

jszwedko:
This field defaults to 300 if empty, but we also need to support setting
it to 0 so I chose to use a *int pointer.

I stumbled upon #56 which discussed the use of pointers to support optional fields which seemed to be leaning towards the approach I took here, but let me know what you think.

`TestCreateAndDeleteMonitor` is failing locally for me, showing a diff including:

```
                        -  Handle: (string) "",
                        -  Id: (int) 0,
                        -  Name: (string) ""
                        +  Email: (string) (len=30) "jesse.szwedko@getbraintree.com",
                        +  Handle: (string) (len=30) "jesse.szwedko@getbraintree.com",
                        +  Id: (int) 222578,
                        +  Name: (string) (len=13) "Jesse Szwedko"
```

I can dig in more, but this also fails on master (first including the changes from #81) so I'm hoping I'm just doing something wrong when running the tests.

### DIFF
--- a/convert_types.go
+++ b/convert_types.go
@@ -1,0 +1,8 @@
+package datadog
+
+func Int(val interface{}) *int {
+	if ival, ok := val.(int); ok {
+		return &ival
+	}
+	return nil
+}
--- a/convert_types.go
+++ b/convert_types.go
@@ -1,8 +1,15 @@
 package datadog
 
-func Int(val interface{}) *int {
-	if ival, ok := val.(int); ok {
-		return &ival
+// Int is a helper routine that allocates a new int value
+// to store v and returns a pointer to it.
+func Int(v int) *int { return &v }
+
+// GetInt is a helper routine that returns a boolean representing
+// if a value was set, and if so, dereferences the pointer to it.
+func GetInt(v *int) (int, bool) {
+	if v != nil {
+		return *v, true
 	}
-	return nil
+
+	return 0, false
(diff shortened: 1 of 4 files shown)


ojongerius:
Nice one! Your implementation matches my preferred solution to #56. The acceptance test were failing, which was raised fixed in #73 / #85.

convert_types.go
```diff
@@ -0,0 +1,8 @@
+package datadog
+
+func Int(val interface{}) *int {
```
ojongerius: I originally started coding up the same implementation, but if your function is named Int, has as single purpose returning a pointer to an int (or it returns nil), why not change the signature to only accept an int like so?:

```go
func Int(v int) *int { return &v }
```

(See https://github.com/zorkian/go-datadog-api/pull/83/files#diff-01127be9ed6b6352616a2f1c10bc3540R29 )

jszwedko: :+1: I tend to agree. I'd borrowed the implementation from `aws/aws-sdk-go` (which I meant to call out in the PR description), but I think explicit is better.

jszwedko: Borrowed your implementation in c08c4994b3d8a0ec40ab35e94a583531de951f5b

ojongerius:
Thanks @jszwedko!

PR #83, opened 2017-01-27
WIP GH-56 Refactor to use pointers for all fields.

ojongerius:
* Refactors fields to be pointers 
* Adds a few setters for common types.

Needs work to let acceptance tests `TestCreateAndDeleteMonitor` and `TestCreateAndDeleteDowntime` pass.

Needs discussion of getters next to the current setters.


ojongerius:
@zorkian feel free to push changes to this branch, I've pushed it here with that in mind.

ojongerius:
Made all acceptance tests pass too.

dashboards.go
```diff
@@ -66,48 +66,48 @@ type Graph struct {
...
-		Autoscale  bool   `json:"austoscale,omitempty"`
-		TextAlign  string `json:"text_align,omitempty"`
-		Precision  string `json:"precision,omitempty"`
-		CustomUnit string `json:"custom_unit,omitempty"`
+		Autoscale  *bool   `json:"austoscale,omitempty"`
+		TextAlign  *string `json:"text_align,omitempty"`
+		Precision  *string `json:"precision,omitempty"`
+		CustomUnit *string `json:"custom_unit,omitempty"`
 
-		// For hostnamp type graphs
```
zorkian: Nice :)

alerts.go
```diff
@@ -15,14 +15,14 @@ import (
...
 type Alert struct {
-	Id           int    `json:"id,omitempty"`
-	Creator      int    `json:"creator,omitempty"`
-	Query        string `json:"query,omitempty"`
-	Name         string `json:"name,omitempty"`
-	Message      string `json:"message,omitempty"`
-	Silenced     bool   `json:"silenced,omitempty"`
-	NotifyNoData bool   `json:"notify_no_data,omitempty"`
-	State        string `json:"state,omitempty"`
+	Id           *int    `json:"id,omitempty"`
```
zorkian: This is unrelated to your changes here but I'm noticing that it looks like inconsistent usage of `omitempty`. The `Alert` struct uses it on every member but `Comment` only uses it on optional members.

helper_test.go
```diff
@@ -0,0 +1,112 @@
+/*
+ * Datadog API for Go
+ *
+ * Please see the included LICENSE file for licensing information.
+ *
+ * Copyright 2016 by authors and contributors.
```
zorkian: 2017?

helper_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"testing"
+
+	"encoding/json"
+	"github.com/stretchr/testify/assert"
+	"github.com/zorkian/go-datadog-api"
+)
+
+func TestBool(t *testing.T) {
```
zorkian: Did you want this to do anything?

helpers.go
```diff
@@ -0,0 +1,67 @@
+/*
+ * Datadog API for Go
+ *
+ * Please see the included LICENSE file for licensing information.
+ *
+ * Copyright 2016 by authors and contributors.
```
zorkian: 2017?

helpers.go
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package datadog
+
+import "encoding/json"
+
+// Bool is a helper routine that allocates a new bool value
+// to store v and returns a pointer to it.
+func Bool(v bool) *bool { return &v }
```
zorkian: Note to self: out of curiosity, does the compiler inline this and go straight to heap? Or does this get passed in on the stack, then heap allocate, then copy, then return pointer? /curiousminds

helpers.go
```diff
@@ -0,0 +1,67 @@
...
+
+// GetBool is a helper routine that returns a boolean representing
+// if a value was set, and if so, dereferences the pointer to it.
+func GetBool(v *bool) (b bool, ok bool) {
+	if v != nil {
+		return *v, true
+	}
+
+	return b, ok
```
zorkian: I know that semantically this returns the zero-value for these which is equivalent to `return false, false` but it seems a bit non-standard to do it this way. What do you think about explicit `return false, false` here?

Or, I could possibly see `return b, false` but at least the `ok` should be defined?

integration/dashboards_test.go
```diff
@@ -151,66 +151,65 @@ func cleanUpDashboard(t *testing.T, id int) {
...
+	graph := datadog.Graph{Title: datadog.String("Mandatory graph 2"), Definition: graphDefinition}
 	graphs := []datadog.Graph{}
 	graphs = append(graphs, graph)
 	return graphs
 }
 
 func assertDashboardEquals(t *testing.T, actual, expected *datadog.Dashboard) {
-	if actual.Title != expected.Title {
-		t.Errorf("Dashboard title does not match: %s != %s", actual.Title, expected.Title)
+	if *actual.Title != *expected.Title {
```
zorkian: Do we want to be using helpers here? (And elsewhere where we just dereference pointers?) See later discussion on helpers.

integration/downtime_test.go
```diff
@@ -10,35 +10,36 @@ func init() {
...
-func TestCreateAndDeleteDowntime(t *testing.T) {
+func TestDowntimeCreateAndDelete(t *testing.T) {
 	expected := getTestDowntime()
 	// create the downtime and compare it
 	actual := createTestDowntime(t)
-	defer cleanUpDowntime(t, actual.Id)
+	defer cleanUpDowntime(t, *actual.Id)
 
-	// Set ID of our original struct to zero we we can easily compare the results
+	// Set ID of our original struct to zero so we can easily compare the results
```
zorkian: ITT comments are hard. Thanks for fixing!

integration/dashboards_test.go
ojongerius: My initial take on the PR was to see if we could agreement on pointers and setters, and I added the getters afterwards.

At the same rate with tests we know what can expect so it's safe to do this.

helpers.go
ojongerius: Agreed, and happy to make that change.

integration/downtime_test.go
ojongerius: 👍

helpers.go
ojongerius: 2017 already indeed..

helper_test.go
ojongerius: Nope! Thanks for spotting it.

helpers.go
ojongerius: I'm not sure, that's a good question and I'd like to find out!

helper_test.go
ojongerius: Yup!

alerts.go
ojongerius: True, I'll add omitempty to all fields 🤞

helpers.go
ojongerius: The `-m` flag to the rescue; it looks like the first: inline and straight to the heap:

```go
 go build -gcflags=-m helpers.go 
# command-line-arguments
<SNIP>
./helpers.go:15: can inline Bool
./helpers.go:15: &v escapes to heap
./helpers.go:15: moved to heap: v
<SNIP>
```

ojongerius:
> Should we switch from int to json.Number everywhere? In most cases it doesn't matter but it feels like it might be useful to standardize instead of having to guess/worry about getting it wrong.

I agree we should be consistent. I'll keep off doing that until we are sure what direction we are going.

> For helpers, what do you think about the way protobufs create GetFoo helpers for each member and handles returning the language-default rather than a possibly-nil which is right most of the time, and if someone really wants to see if it's nil they can inspect the member. (The comma ok idiom is very verbose.) I'm interested in keeping the API simple to use, but correct. Any thoughts?

I don't have experience with protobufs, can you share a code snippet on how a client would use it, and how we would code those getters? Definitely a fan of finding a clean way of not having to use the `comma ok` idiom too.

> I mentioned in one place but there is inconsistency around the deployment of omitempty or not. Maybe we should just put it everywhere and leave it up to the user to make sure they set the fields they need to set. Since the DD API will error if they do it wrong, it saves us from getting mismatched expectations -- i.e. if DD decides to make something optional and the library is forcing it, someone has to patch it before they can use it which is a bummer.

Agreed and implemented.

> And as an aside, I'm going to do an experiment -- protobufs actually have a marshal-to-JSON option. I'm wondering if I write a proto definition for these API methods if it would Just Work or if the JSON it outputs can't be made compatible with DD's API? Worth a shot anyway.

Interested to see how you go! Will Norris started to port go-github to use protobufs but ended up reverting that and cherry picking some of protobufs functionality. See this comment https://github.com/google/go-github/issues/19#issuecomment-22976895 .

helpers.go
zorkian: Nice! Well, that makes me happy. Thanks for checking. :)

integration/dashboards_test.go
zorkian: Makes sense. Well, I'm agreed on using the methods like `Int` and `String` etc. Although if we do go with the `json.Number` route we should make it `Number` instead? (Oh ugh, nevermind, I forget Go can't overload arguments... oh wait, I guess we could make it take an `interface{}` and unbox it ourselves??)

zorkian:
> I don't have experience with protobufs, can you share a code snippet on how a client would use it, and how we would code those getters? Definitely a fan of finding a clean way of not having to use the comma ok idiom too.

I was thinking like this: https://play.golang.org/p/sh2M9cIs-2

That'd be pretty annoying to have to build ourselves though, but maybe some awk/sed/scripting can generate it appropriately. Or, actually, something like this:

https://play.golang.org/p/Cq9dWFqAd0

If we throw that in some autogen/ package it can import all the structs and then dump out the appropriate methods. If we were really fancy it could just automatically rewrite the source files or something. (Or possibly some magic with `gofmt -r` but I don't know how advanced that can get.)

Okay sorry, I've gone off a bit into the weeds here. I think easy-to-use is a noble goal if we can achieve it, but I'm not asserting that having `Get...` and `Has...` methods is the way to go. I kind of like the simplicity but it is a lot of work to maintain.

ojongerius:
Am reaching out to Datadog folks if we could get a service descriptor from them.

Generating the whole library a la https://github.com/google/google-api-go-client or https://github.com/aws/aws-sdk-go would be an direction I'd be interested in to work on.

@zorkian thoughts?

ojongerius:
Datadog is not going to work on publishing their API any time soon 😞

On a more positive note I've coded something up that generates HasXx, GetXx and SetXx. Using it would look like:

```go
	m := datadog.Monitor{}
	c := datadog.Creator{}
	
        // Use SetXx:
        c.SetName("Joe Creator")
	m.SetName("Monitor all the things")
	m.SetMessage("Fat electrons in the lines")

	// Use HasXx:
	if m.HasMessage() {
		fmt.Printf("Found message %s\n", m.GetMessage())
	}

	// Or maybe you prefer a tuple with the value and a boolean if it was ever set
        // and use GetXxOk
	if v, ok := m.GetMessageOk() ; ok {
		fmt.Printf("Found message %s\n", v)
	}
```

I still like having the other helpers around, and keep support for doing:

```go
	// You could still initialize the struct in one swoop
	_ = datadog.Monitor{
		Name: datadog.String("Monitor other things"),
		Creator: &datadog.Creator{
			Name: datadog.String("Joe Creator"),
		},
	}
```

I'll push up some changes to this PR hopefully today, else tomorrow.

ojongerius:
@zorkian thoughts? This PR is getting a little large, and has been out for a while. 

AFAICS this is good to go to v2. Keen to hear your feedback.

cmd/tools/gen-accessors.go
```diff
@@ -0,0 +1,344 @@
...
+		if err := t.dump(); err != nil {
+			log.Fatal(err)
+		}
+	}
+	logf("Done.")
+}
+
+// processAST walks source code files, looks for struct definitions, and calls appropriate helper functions
+// to add generate accessors for type fields are star expressions -ie: pointers
+func (t *templateData) processAST(f *ast.File) error {
```
ojongerius: This currently only goes one level deep. If we want to support nested structs this will needs refactoring. Although we have nested structs at the moment, I'm not super keen on them.

My preference at the moment is to get all this merged into v2, and if we agree on the direction, support accessors for nested structs in a separate PR. It's not much work but this PR is a big of a change as it is.

AFAIC, before we make v2 the default branch we can safely to this and test with v2.

@zorkian thoughts?

zorkian: This is a much fancier approach than the one I had for sure! I think ATM having top-level only is fine, and we can see how things develop.

ojongerius:
I did run the generator, not a 100% if it contains all the changes. All generated accessors are in `datadog-accessors.go` which did not show up as a diff as it is a new file. We could also chose to have the accessors closer to the code structs. Up to us.

Thanks for the review @zorkian! Yes code regeneration will have to be run manually, I imagine we could be smart about this and have a make target that verifies that code has to be regenerated. For now I will merge this and we can get fancier later.

I might have a stab at updating Terraform to use the new version of the library and see if I run into issues. We could also entertain adding instructions on how to use the new branch to master.

PR #87, opened 2017-02-15
Allow null for new_host_delay

jszwedko:
Otherwise there isn't a great way to tell the monitor to fall back to the Datadog default (currently 300s) as Datadog treats the absence of the key different from the presence of the key with `null` as the value.

### DIFF
--- a/monitors.go
+++ b/monitors.go
@@ -42,7 +42,7 @@ type Options struct {
 	NoDataTimeframe   NoDataTimeframe `json:"no_data_timeframe,omitempty"`
 	NotifyAudit       bool            `json:"notify_audit,omitempty"`
 	NotifyNoData      bool            `json:"notify_no_data,omitempty"`
-	NewHostDelay      *int            `json:"new_host_delay,omitempty"`
+	NewHostDelay      *int            `json:"new_host_delay"`
 	RenotifyInterval  int             `json:"renotify_interval,omitempty"`
 	Silenced          map[string]int  `json:"silenced,omitempty"`
 	TimeoutH          int             `json:"timeout_h,omitempty"`


ojongerius:
Hi @jszwedko,

AFAICS using the `omitempty` option in combination with *int works as advertised. Can you help me reproduce this?

I'm under the impression the following code asserts that *not* setting NewHostDelay results in the default value of 300, and setting it to a non default value results in that value being returned. Happy to be proven wrong.

```go
	fmt.Println("Monitor without host delay set")

	// Options with NewHostDelay *not* set:
	o := datadog.Options{
		NotifyNoData:    true,
		NoDataTimeframe: 60,
		Silenced:        map[string]int{},
	}

	m := datadog.Monitor{
		Message: "Test message",
		Query:   "avg(last_15m):avg:system.disk.in_use{*} by {host,device} > 0.8",
		Name:    "Test monitor",
		Options: o,
		Type:    "metric alert",
		Tags:    make([]string, 0),
	}

	monitor, err := client.CreateMonitor(&m)
	if err != nil {
		t.Fatalf("Creating a monitor failed when it shouldn't: %s", err)
	}

	defer cleanUpMonitor(t, monitor.Id)

	// Assert that not setting it results in the (currrently) default value fo 300:
	assert.Equal(t, 300, *monitor.Options.NewHostDelay)

	fmt.Println("Monitor with host delay set")

	// Set NewHostDelay to a non default number
	hostDelayVal := 666
	m.Options.NewHostDelay = &hostDelayVal

	monitor, err = client.CreateMonitor(&m)

	defer cleanUpMonitor(t, monitor.Id)

	// Assert that our request was honored:
	assert.Equal(t, 666, *monitor.Options.NewHostDelay)
```

And it passes.

jszwedko:
Hi @ojongerius,

Apologies, I wasn't clear in my original description, the issue is that if you set a value for `NewHostDelay`, but then want to remove it to allow Datadog to default back to `300`, you need to include `"new_host_delay": null` in the `PUT` options. Here is a modification of your test case that should hopefully make this clearer:

```golang
        fmt.Println("Monitor without host delay set")

        // Options with NewHostDelay *not* set:
        o := datadog.Options{
                NotifyNoData:    true,
                NoDataTimeframe: 60,
                NewHostDelay:    datadog.Int(666),
                Silenced:        map[string]int{},
        }

        m := datadog.Monitor{
                Message: "Test message",
                Query:   "avg(last_15m):avg:system.disk.in_use{*} by {host,device} > 0.8",
                Name:    "Test monitor",
                Options: o,
                Type:    "metric alert",
                Tags:    make([]string, 0),
        }

        monitor, err := client.CreateMonitor(&m)
        if err != nil {
                t.Fatalf("Creating a monitor failed when it shouldn't: %s", err)
        }
        defer cleanUpMonitor(t, monitor.Id)

        if err := client.UpdateMonitor(monitor); err != nil {
                t.Fatalf("Updating a monitor failed when it shouldn't: %s", err)
        }

        actual, err := client.GetMonitor(monitor.Id)
        if err != nil {
                t.Fatalf("Retrieving a monitor failed when it shouldn't: %s", err)
        }

        // Assert that not setting it results in the default value fo 300:
        assert.Equal(t, 300, *actual.Options.NewHostDelay)
```

Thank you for taking a look at this!

ojongerius:
I was able to run the following test successfully, against the v2 branch, which use the a pointer and contains `omittempy`:

```go
func TestNewHostDelayMonitorReply(t *testing.T) {
	fmt.Println("Monitor with host delay set")

	// Options with NewHostDelay set to 666:
	o := datadog.Options{
		NotifyNoData:    datadog.Bool(true),
		NoDataTimeframe: 60,
		Silenced:        map[string]int{},
		NewHostDelay:    datadog.Int(666),
	}

	m := datadog.Monitor{
		Message: datadog.String("Test message"),
		Query:   datadog.String("avg(last_15m):avg:system.disk.in_use{*} by {host,device} > 0.8"),
		Name:    datadog.String("Test monitor"),
		Options: &o,
		Type:    datadog.String("metric alert"),
		Tags:    make([]string, 0),
	}

	monitor, err := client.CreateMonitor(&m)
	if err != nil {
		t.Fatalf("Creating a monitor failed when it shouldn't: %s", err)
	}

	defer cleanUpMonitor(t, monitor.GetId())

	// Assert that it's set to 666:
	assert.Equal(t, 666, *monitor.Options.NewHostDelay)

	fmt.Println("Monitor without host delay set")

	//Options with NewHostDelay *not* set:
	updatedOptions := datadog.Options{
		NotifyNoData:    datadog.Bool(true),
		NoDataTimeframe: 60,
		Silenced:        map[string]int{},
	}

	monitor.Options = &updatedOptions

	if err = client.UpdateMonitor(monitor) ; err != nil {
		t.Fatalf("Creating a monitor failed when it shouldn't: %s", err)
	}

	monitor, err = client.GetMonitor(monitor.GetId())
	if err != nil {
		t.Fatalf("Creating a monitor failed when it shouldn't: %s", err)
	}

        // Assert value is back to the default setting
	assert.Equal(t, 300, monitor.Options.GetNewHostDelay())
}
```

In the code example you share, if I read it correctly, you create a monitor, do an update without changing the options, so I'm not surprised that fails.

Hope this helps, happy to be proven wrong :)

jszwedko:
Doh, yep, thank you @ojongerius . Apologies for the confusion; it does appear to have the effect I expected even with `omitempty`.

ojongerius:
No worries 👍

Issue #88, opened 2017-02-17
Align with Datadog on their fork


ojongerius:
I noticed their is a fork at https://github.com/DataDog/go-datadog-api, and I'm curious if Datadog contacted you, @zorkian?

I noticed https://github.com/ufoot commit a change, and I'm wondering what the vision is here If the code of up and downstream start to deviate that would be a shame.

zorkian:
No, I haven't had any communication with them. PS I will get back to you on the other PR, it's looking good so far. I'm mid-review but haven't finished yet, sorry!

ojongerius:
Interesting, thanks. Good to know you are on the PR :)

PR #89, opened 2017-02-17
[series] add support for metrics unit

LeoCavaille:
When submitting a datapoint, one can also submit a "unit" information that will be used to update metrics metadata and eventually what's displayed on the UI.

Related docs:
* http://docs.datadoghq.com/units/
* https://help.datadoghq.com/hc/en-us/articles/204580889-How-can-I-set-up-custom-units-for-custom-metrics-

To also respond to #88, we initially forked to hack on top of the lib for an internal project at Datadog and we had plans to submit upstream but just did not follow through. We are really thankful for all the great work putting together this lib, we are heavy users now :bowing_man: . Once merged, we can pin the newer version upstream in our codebase and remove the fork :)

### DIFF
--- a/series.go
+++ b/series.go
@@ -22,6 +22,7 @@ type Metric struct {
 	Type   string      `json:"type,omitempty"`
 	Host   string      `json:"host,omitempty"`
 	Tags   []string    `json:"tags,omitempty"`
+	Unit   string	   `json:"unit,omitempty"`
 }
 
 // Series represents a collection of data points we get when we query for timeseries data


ojongerius:
Thanks for pushing this upstream @LeoCavaille! Go to hear you are finding the library useful.

zorkian:
Oh, that's great to hear! Yeah, I'm personally a big fan of the Datadog
service and use it daily, so. Thank you!

PR #91, opened 2017-02-18
Don't use doRequestWithRetries with PUT requests

bshelton229:
This bug was discovered here:

https://github.com/hashicorp/terraform/issues/11106

When updating an existing monitor with incorrect data a retrying request is made rather than trying the PUT and returning the correct error immediately.


zorkian:
Nice catch. Yeah, I think for things that potentially modify remote state, not retrying is the correct behavior.

ojongerius:
👍

Issue #76, opened 2017-01-09
POST errors leak credentials


grubernaut:
Discovered via: https://github.com/hashicorp/terraform/issues/11107

When a POST request comes into the DataDog API via: https://github.com/hashicorp/terraform/blob/master/builtin/providers/datadog/resource_datadog_monitor.go#L227, the `go-datadog-api` calls `*http.Client.Do()`. 

Due to the fact that the API Key and Application Key are provided in the URI, and the URI is included inside the returned Error; this leads to leaking credentials on the CLI where the error is returned. 

cc: @keymon who reported the issue to Terraform.

PR #92, opened 2017-02-18
doRequestWithRetries doesn't work correctly when req.Body is not nil

bshelton229:
I still think not retrying PUT requests is the right thing to do, but this was responsible for the strange errors seen while retrying the PUT update of a monitor seen in #91.

`HttpClient.Do` was returning errors on retries when the request contained a `Body`. When the http request contains a body it seems it cannot be retried because the `req.Body` is closed after the first try. After the first try the request always errors out because of the empty body not matching the ContentLength header. The issue seems to be the same as this:

https://github.com/mitchellh/goamz/issues/63

#91 may lessen the impact of #76 . As far as I know the only requests that would be sent to datadog which would contain a non nil body would be PUT or POST, which are now not retried. #91 definitely doesn't solve #76 but makes it less likely to be exposed.

This fix is probably not necessary now that PUT isn't retried. I don't think any DELETE requests contain bodies, but it could start a conversation.

I'm not sure what the best approach to sanitizing the error message containing the full request URL would be? I'm going to check with datadog to see if they have an undocumented way to pass credentials via a header.

### DIFF
--- a/request.go
+++ b/request.go
@@ -110,11 +110,25 @@ func (client *Client) doRequestWithRetries(req *http.Request, maxTime time.Durat
 		err  error
 		resp *http.Response
 		bo   = backoff.NewExponentialBackOff()
+		body []byte
 	)
 
 	bo.MaxElapsedTime = maxTime
 
+	// Save the body for retries
+	if req.Body != nil {
+		body, err = ioutil.ReadAll(req.Body)
+		if err != nil {
+			return resp, err
+		}
+	}
+
 	operation := func() error {
+		if body != nil {
+			r := bytes.NewReader(body)
+			req.Body = ioutil.NopCloser(r)
+		}
+
 		resp, err = client.HttpClient.Do(req)
 		if err != nil {
 			return err


ojongerius:
Thanks!

Very much appreciated if you can check with Datadog and/or convince to stop requiring to put credentials in query params.

request.go
```diff
@@ -77,6 +77,9 @@ func (client *Client) doJsonRequest(method, api string,
 		resp, err = client.doRequestWithRetries(req, client.RetryTimeout)
 	}
 	if err != nil {
+		// This will leak datadog credentials due to the error containing
```
ojongerius: Mind adding the related issue id to this comment?

request.go
```diff
@@ -146,12 +149,26 @@ func (client *Client) doRequestWithRetries(req *http.Request, maxTime time.Durat
...
+			return resp, err
+		}
+	}
+
 	err = backoff.Retry(func() error {
+		r := bytes.NewReader(body)
+		req.Body = ioutil.NopCloser(r)
 		resp, err = client.HttpClient.Do(req)
 		if err != nil {
+			// This will leak datadog credentials due to the error containing
```
ojongerius: Maybe put this information (plus related issue) a little higher so we don't need to duplicate it?

ojongerius:
When reviewing I #94 for which I've raised #95. Apologies if this causes a merge conflict for you. The PR is against v2 but it needs to land in master too.

bshelton229:
@ojongerius I rebased on top of `v2`. The changes in #95 are great! Could we backport those to v1 to get them into the terraform provider, or are there plans to update that to use v2 soon? I ran into a case where that will help. If you GET a deleted monitor, that will respond legitimately with a 404, which will be retried until the timeout before the monitor will be re-created by terraform.

I removed the comments about api credential leakage. It's probably easier to deal with that separately. I did contact datadog. They've had requests for authentication via HTTP headers in the past but it doesn't seem like it's an active project at this point. I let them know we'd really like to see that happen and will keep after it.

Thanks for this library! This in conjunction with the terraform datadog provider has provided huge benefits for us.

ojongerius:
@bshelton229, appreciated, it's great to hear people are enjoying the work.  Let's discuss the API credential leakage in #76, thanks for the investigation!
 
The v2 branch has made it into Terraform via https://github.com/hashicorp/terraform/pull/12098.

Issue #94, opened 2017-02-19
Retry logic is flawed


ojongerius:
Currently acceptance tests take a very long time, one test takes ~90 seconds:

```bash
github.com/zorkian/go-datadog-api  v2 ✗                                            2h27m ⚑ ◒
▶ time make testacc
go test integration/* -v  -timeout 90m
=== RUN   TestInvalidAuth
--- PASS: TestInvalidAuth (1.27s)
=== RUN   TestValidAuth
--- PASS: TestValidAuth (0.31s)
=== RUN   TestDashboardCreateAndDelete
--- PASS: TestDashboardCreateAndDelete (86.53s)
=== RUN   TestDashboardCreateAndDeleteAdvancesTimeseries
--- PASS: TestDashboardCreateAndDeleteAdvancesTimeseries (86.80s)
=== RUN   TestDashboardUpdate
--- PASS: TestDashboardUpdate (65.49s)
=== RUN   TestDashboardGet
<SNIP>
```

What happened: github.com/cenkalti/backoff will retry every error. This resulted in all GETs (to confirm resources have been deleted) to retry until the max amount of time had passed.

After fixing up the backoff logic the *total* running time is now ~90 seconds:

```bash
--- PASS: TestInvalidAuth (1.21s)
=== RUN   TestValidAuth
--- PASS: TestValidAuth (0.26s)
=== RUN   TestDashboardCreateAndDelete
--- PASS: TestDashboardCreateAndDelete (1.45s)
=== RUN   TestDashboardCreateAndDeleteAdvancesTimeseries
--- PASS: TestDashboardCreateAndDeleteAdvancesTimeseries (1.40s)
=== RUN   TestDashboardUpdate
<SNIP>
=== RUN   TestScreenboardGet
--- PASS: TestScreenboardGet (2.67s)
=== RUN   TestUserCreateAndDelete
--- PASS: TestUserCreateAndDelete (2.02s)
PASS
ok      command-line-arguments  90.757s
make testacc  1.94s user 0.34s system 2% cpu 1:31.91 total
```

- [ ] Cherry pick c4db4135faf89a3d48c5ace9bff59d3dd9cbdcde into master if reviewed ok

ojongerius:
What happened: backoff will retry *every* error. This resulted in all GETs (to confirm resources have been deleted) to retry until the max amount of time had passed.

ojongerius:
I'm going to refrain from pulling this into v1, as we'll be cutting over soon anyway.

PR #95, opened 2017-02-19
Backoff

ojongerius:
Addresses #94

### DIFF
--- a/request.go
+++ b/request.go
@@ -128,7 +128,7 @@ func (client *Client) doRequestWithRetries(req *http.Request, maxTime time.Durat
 			return nil
 		}
 
-		return fmt.Errorf("Received HTTP status code %d\n", resp.StatusCode)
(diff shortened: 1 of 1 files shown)


request.go
```diff
@@ -105,26 +104,34 @@ func (client *Client) doJsonRequest(method, api string,
...
+			// 2xx all done
+			return nil
+		} else if resp.StatusCode >= 400 && resp.StatusCode < 500 {
+			// 4xx are not retryable
+			return nil
 		}
-		return nil
-	}, bo)
+
+		return fmt.Errorf("Received HTTP status code %d\n", resp.StatusCode)
```
zorkian: Out of curiosity, does the `\n` do anything in particular with a formatted error string? It seems like I'd probably not insert one and let the caller deal with adding one if they're printing out a line.

ojongerius: Good catch! Will remove this.

PR #104, opened 2017-03-06
Add support for yaxis min/max to be "auto"

### LINKED ISSUE #103: Dashboard yaxis assume floats
In dashboard.go, the struct is not quite correct:

```go
type Yaxis struct {
	Min   *float64 `json:"min,omitempty"`
	Max   *float64 `json:"max,omitempty"`
	Scale *string  `json:"scale,omitempty"`
}
```

When the value of min and/or max is set to "auto" from the UI, GetDashboard will fail because demarshalling fails.

yfronto:
* Adds an OptionalFloat64 type with a *float64 and a bool called "auto" to allow for explicit setting of auto in the datadog dashboard UI
* Fixes #103

### DIFF
--- a/dashboards.go
+++ b/dashboards.go
@@ -54,13 +54,14 @@ type GraphEvent struct {
 
 type OptionalFloat64 struct {
 	Value *float64
-	Auto  bool
+	Auto  *bool
 }
 
 func (of *OptionalFloat64) UnmarshalJSON(data []byte) error {
 	s := string(data)
 	if s == "\"auto\"" {
-		of.Auto = true
+		b := true
+		of.Auto = &b
 		of.Value = nil
 	} else if s == "false" {
 		of = nil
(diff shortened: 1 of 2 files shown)


zorkian:
What happens if we re-run the accessor generator? Does it overwrite these updates, or does it leave unchanged methods that already exist?

yfronto:
My fault, didn't know there was a generator for that. I'll take a look.

yfronto:
Can look at this more later today. It works, but a quick glance shows the generated accessors don't include any info about the auto field. As an aside, it might be worth having CI tests check for diffs on the generator from the commit.

yfronto:
Took another look at this. The problem is that the Auto field and the Value field for the OptionalFloat64 I added aren't related by the generator. So, you can get a dashboard that has a value set, set the auto field, and we won't know how to marshal that (likewise, if auto was set and you set the value). The setters need to also unset the other value.

Not sure what the right approach is here.

PR #109, opened 2017-06-09
add metric metadata

travisjeffery:
- Adds metric metadata ([API docs](http://docs.datadoghq.com/api/#metrics-metadata-get))
- This would allow you to edit metric metadata, for instance, adding a description for a metric to let other people know what the metric means and its values should be.
- Didn't add a test b/c it'd require posting an event and I saw that you don't have tests for posting events so I kept it consistent, let me know and I can fix that

### DIFF
--- a/datadog-accessors.go
+++ b/datadog-accessors.go
@@ -7205,6 +7205,192 @@ func (m *Metric) SetUnit(v string) {
 	m.Unit = &v
 }
 
+// GetDescription returns the Description field if non-nil, zero value otherwise.
+func (m *MetricMetadata) GetDescription() string {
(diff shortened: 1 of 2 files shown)


travisjeffery:
Ping @zorkian - I'd like to use this PR in a PR for Terraform's Datadog provider.

zorkian:
Thanks for the ping -- LGTM! And nice to see the unit test to update the generated APIs is doing good work :)

travisjeffery:
@zorkian can you push an update to gopkg.in/zorkian/go-datadog-api.v2 for this? Thanks

PR #110, opened 2017-06-22
use map[string]string to pass query options for snapshot

selfsk:
* wanted to use graph_def instead of query_metric argument
 * make making snapshots of dashboards much simplier -> GetDashboard() -> Graphs[i].GraphDefinition -> Jsonify() (not in this repo) and query using output JSON

### DIFF
--- a/integration/snapshot_test.go
+++ b/integration/snapshot_test.go
@@ -1,0 +1,59 @@
+package integration
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+	"time"
+)
+
+func Jsonify(v interface{}) (string, error) {
+	if b, err := json.MarshalIndent(v, "", " "); err != nil {
+		return "", err
+	} else {
+		return string(b), nil
+	}
+}
+
+func init() {
+	client = initTest()
+}
+
+func TestSnapshot(t *testing.T) {
+
+	end := time.Now().Unix()
+	start := end - 3600
+
+	query_s := "avg:system.mem.used{*}"
+	url, err := client.Snapshot(query_s, time.Unix(start, 0), time.Unix(end, 0), "")
+
+	if err != nil {
+		t.Fatalf("Couldn't create snapshot for query(%s): %s", query_s, err)
+	}
+
+	fmt.Printf("query snapshot url: %s\n", url)
+}
+
+func TestSnapshotGeneric(t *testing.T) {
+
+	end := time.Now().Unix()
+	start := end - 3600
+
+	// create new graph def
+	graphs := createCustomGraph()
+
+	graph_def, err := Jsonify(graphs[0].Definition)
+	if err != nil {
+		t.Fatalf("Couldn't create graph_def: %s", err)
+	}
+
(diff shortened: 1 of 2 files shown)


zorkian:
This breaks the API for this method. How would you feel about making a second method that uses the new calling conventions (which I agree are superior)? Otherwise I'd have to bump to v3 for semantic versioning compliance which I'd prefer not to do for a one-method change, I think.

zorkian:
Also, sorry for the delay in review -- have been traveling.

selfsk:
make sense, I'll update PR then (or you want new one?)

selfsk:
PR has been updated

nyanshak:
@selfsk Thanks for putting this together. I was just about to put an issue in about the same thing.

What do you think about using the [graphing json grammar](https://docs.datadoghq.com/graphingjson/#grammar) to make more well-defined types than using a map of string to string?

@zorkian cc

selfsk:
@nyanshak well-defined types, not sure what exactly you mean by this. But you can use `graph_def: value`, to get complex graphs. Like I do in test - https://github.com/zorkian/go-datadog-api/pull/110/files/17834f8c208e0eb1533c3491d5874374cfc35fda#diff-10662fbcd9ae611dd9fe16e1cfd2f8c9R50

nyanshak:
What I mean by well-defined types is changing the function signature to something like this:

```
func Snapshot(datadog.Graph, start, end time.Time, eventQuery string) (string, error) {
```
By doing this it's more clear what type to pass to the function than `options map[string]string`.

With the current code, this would compile:
```
options := map[string]string{
    "Stacked": "17",
    "Autoscale": "17",
    "IncludeNoMetricHosts": "17",
    "Events": "{}",
}

// assume client is an existing datadog client with valid creds
client.Snapshot(options, start, end, eventQuery)
```
However you'd get a runtime error upon trying to actually call Snapshot. Whereas with datadog.Graph as the expected type (for example), you would be forced to make a valid graph type.

@selfsk

selfsk:
@nyanshak I see. Idea behind map[string]string is that you can use different param:value options for request (https://docs.datadoghq.com/api/#graphs)

```
options := map[string]string{
   "metric_query": "...",
   "title": "...."
}
```

Or
```
options := map[string]string{
   "graph_def": "{....}",
   "title": "...."
}
```

It's more generic option, I can create another function like `SnapshotByGraphDef(d Definition)`, where I will Jsonify Definition instance and pass it as 'graph_def'. But you always can do it yourself :)

I hope that make sense

zorkian:
(Sorry again, was out of town for the Eclipse. I'm back now for the rest of the year except a few long weekends...)

Given that this is named `SnapshotGeneric` I think it's fine to take the map for now.

I like the idea of being able to ask for Snapshots with well-defined types, in that case tho I'd picture that being a method on the `datadog.Graph` type? cc @nyanshak I'm generally in favor of minimalist "library as API interface", but I want it to also do an appropriate amount of heavy lifting so every user of the library doesn't have to write the same kind of boilerplate. Any thoughts?

PR #117, opened 2017-08-30
metrics metadata: fix statsd interval to be an int

travisjeffery:
Last fix. This API has been hell.

Thanks and sorry for the trouble, @zorkian.


zorkian:
@travisjeffery Merged & release `v2.5` made.

PR #118, opened 2017-09-16
Improved API URI generation logic

tariq1890:


tariq1890:
@zorkian The test cases fail since Go 1.6 does not support t.Run(). If you need to support Go 1.6, I can rewrite the test cases in the old style.

zorkian:
Hey @travisjeffery @ojongerius any opinions on 1.6 support for your needs, as you both use this module?

I personally don't think we need to, as go 1.7 has been out for over a year.

zorkian:
@tariq1890 If they think it's fine to remove 1.6 support, please update this diff to remove it from the travis test file. If we don't hear back in a day or two we should just proceed with removing 1.6 support. Thanks!

travisjeffery:
Removing 1.6 support sounds good to me. 👍

tariq1890:
Cool! I'll wait on @ojongerius  then. @zorkian I see that you have requested changes, but I am not able to see any review comments.

zorkian:
I think we don't need to block on @ojongerius, we've gotten one ACK and it SGTM so let's go for it. Can you update the diff to remove 1.6 from the Travis configuration so the tests pass? Then I will merge it.

tariq1890:
@zorkian Done!  It's passing now

zorkian:
Thanks! Merged!

ojongerius:
No problem for me, thanks for the work @tariq1890 !

PR #122, opened 2017-10-22
update gen-accessors.go to generate more accurate docs

dustinlindquist:
when working with the datadog api via the [terraform-provider](https://github.com/terraform-providers/terraform-provider-datadog) I saw this small doc bug.


zorkian:
Awesome, thank you!

PR #124, opened 2017-10-24
Redact api and app keys from errors

### LINKED ISSUE #76: POST errors leak credentials

nyanshak:
Fix #76.

I don't know if this is the right way to fix this but I took a stab at it.

### DIFF
--- a/request.go
+++ b/request.go
@@ -17,6 +17,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/cenkalti/backoff"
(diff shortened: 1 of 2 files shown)


request.go
```diff
@@ -41,9 +42,25 @@ func (client *Client) uriForAPI(api string) (string, error) {
 }
 
 // doJsonRequest is the simplest type of request: a method on a URI that returns
-// some JSON result which we unmarshal into the passed interface.
+// some JSON result which we unmarshal into the passed interface. It wraps
+// doJsonRequestUnsafe to redact api and application keys from errors.
 func (client *Client) doJsonRequest(method, api string,
 	reqbody, out interface{}) error {
+	err := client.doJsonRequestUnsafe(method, api, reqbody, out)
```
ojongerius: A nit picking alternative would be:

```go
   if err := client.doJsonRequestUnsafe(method, api, reqbody, out) ; err != nil {
        errString := strings.Replace(err.Error(), client.apiKey, "redacted", -1)
        errString = strings.Replace(errString, client.appKey, "redacted", -1)
        return fmt.Errorf(errString)
    }
```

request.go
```diff
@@ -41,9 +42,25 @@ func (client *Client) uriForAPI(api string) (string, error) {
...
+		errString = strings.Replace(errString, client.apiKey, "redacted", -1)
+		errString = strings.Replace(errString, client.appKey, "redacted", -1)
+		return fmt.Errorf("%s", errString)
+	}
+	return nil
+}
+
+// doJsonRequestUnsafe is the simplest type of request: a method on a URI that returns
+// some JSON result which we unmarshal into the passed interface.
+func (client *Client) doJsonRequestUnsafe(method, api string,
```
ojongerius: Would `doJsonRequestUnredacted` be a more informative name?

request.go
nyanshak: Will update this ^

ojongerius:
@nyanshak do you have time to add a test to make sure this issue does not regress? Making this change would be a  good time to add it.

nyanshak:
Rebased off master to resolve conflicts, also added the tests you mentioned @ojongerius. What do you think? :)

zorkian:
Awesome, thanks for adding the tests! LGTM!

PR #127, opened 2017-10-24
Replace `is_admin` with `access_role`

### LINKED ISSUE #126: Support `access_role` field for user object
Proposal: Replace `is_admin` with `access_role`.

Reasons:

There are currently 3 types of users `ro` (read-only), `st` (standard), and `adm` (admin). The library as-is (with `is_admin` field) only supports `st` (`is_admin` set to false) and `adm` (`is_admin` set to true). This means that read-only users will automatically be changed to standard users when this library is used to update a user, even if that was not the intention.

Why not keep both `is_admin` and `access_role`?

This results in confusing behavior, because it's unclear to users what the behavior is if I set `is_admin` to true and `access_role` to `st`. Do I get a standard user that somehow has admin access? Do I get a standard user? Do I get an admin user? The correct answer is that `access_role` takes precedence and `is_admin` will be ignored if they conflict.

```
| is_admin | access_role | resulting user |
| false    | ro          | ro             |
| true     | ro          | ro             |
| false    | st          | st             |
| true     | st          | st             |
| false    | adm         | adm            |
| true     | adm         | adm            |
```

@zorkian Let me know what you think about the proposal.

nyanshak:
Fix #126

### DIFF
--- a/datadog-accessors.go
+++ b/datadog-accessors.go
@@ -11700,6 +11700,37 @@ func (u *User) SetHandle(v string) {
 	u.Handle = &v
 }
 
+// GetIsAdmin returns the IsAdmin field if non-nil, zero value otherwise.
+func (u *User) GetIsAdmin() bool {
+	if u == nil || u.IsAdmin == nil {
+		return false
+	}
+	return *u.IsAdmin
+}
+
+// GetOkIsAdmin returns a tuple with the IsAdmin field if it's non-nil, zero value otherwise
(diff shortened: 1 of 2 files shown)


zorkian:
Oh no, I'm so sorry this got missed! I think I missed that you had updated the PR, I'm so sorry about that. If you don't get a response within a few days on these please feel free to bump the issue/PR!

PR #128, opened 2017-10-26
change http-method GET to POST on api.Screenboard#share

haohcraft:
For the CSRF concern, we have changed GET to POST for screenboard:share API.

### DIFF
--- a/screenboards.go
+++ b/screenboards.go
@@ -107,7 +107,7 @@ type ScreenShareResponse struct {
 
 // ShareScreenboard shares an existing screenboard, it takes and updates ScreenShareResponse
 func (client *Client) ShareScreenboard(id int, response *ScreenShareResponse) error {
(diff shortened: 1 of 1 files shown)


nyanshak:
What's the CSRF concern here? Is there an issue, datadog support page, or some other place describing it?

haohcraft:
Since this get endpoint is to mute the sharable state, it's better to use POST instead.

ojongerius:
@haohcraft thanks for the PR! According to the docs at https://docs.datadoghq.com/api/?lang=console#screenboards sharing takes  a GET, creation a POST and updating a PUT, -as you would expect from a REST endpoint.

Could you please elaborate on what gave you the motivation to make this change? I'm interested to understand your concern here.

haohcraft:
According to https://docs.datadoghq.com/api/?lang=console#screenboards-share, it should be a POST now.

ojongerius:
Thanks @haohcraft!

PR #129, opened 2017-11-06
Accept "null" in monitor NoDataTimeframe

dtan4:
## Why

If the monitor is configured _not_ to notify data missing, GetMonitor API returns `null` in `no_data_timeframe` field.
Therefore `NoDataTimeframe.UnmarshalJSON` tries to parse `"null"` as `Int`, then it returns error.

### how to reproduce

```go
package main

import (
	"os"

	"github.com/k0kubun/pp"
	"github.com/zorkian/go-datadog-api"
)

const (
	monitorID = 123456 // your monitor ID which does not notify data missing
)

func main() {
	apiKey, appKey := os.Getenv("DATADOG_API_KEY"), os.Getenv("DATADOG_APP_KEY")
	client := datadog.NewClient(apiKey, appKey)

	mon, err := client.GetMonitor(monitorID)
	if err != nil {
		panic(err)
	}

	pp.Println(mon)
}
```

```sh-session
$ go run hoge.go
panic: strconv.ParseInt: parsing "null": invalid syntax

goroutine 1 [running]:
main.main()
        /Users/dtan4/src/github.com/dtan4/terraforming-datadog/hoge.go:20 +0x1e1
exit status 2
```

## What

Accept `nil` in `no_data_timeframe` and return empty value.

### DIFF
--- a/integration/monitors_test.go
+++ b/integration/monitors_test.go
@@ -70,6 +70,26 @@ func TestMonitorGet(t *testing.T) {
 	}
 }
 
+func TestMonitorGetWithoutNoDataTimeframe(t *testing.T) {
+	monitors, err := client.GetMonitors()
+	if err != nil {
+		t.Fatalf("Retrieving monitors failed when it shouldn't: %s", err)
+	}
+	num := len(monitors)
+
+	monitor := createTestMonitorWithoutNoDataTimeframe(t)
+	defer cleanUpMonitor(t, *monitor.Id)
+
+	monitors, err = client.GetMonitors()
+	if err != nil {
+		t.Fatalf("Retrieving monitors failed when it shouldn't: %s", err)
+	}
+
+	if num+1 != len(monitors) {
+		t.Fatalf("Number of monitors didn't match expected: %d != %d", len(monitors), num+1)
+	}
+}
+
 func TestMonitorMuteUnmute(t *testing.T) {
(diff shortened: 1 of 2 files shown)


philwhln:
I'm hitting this issue too. This change makes sense to me.

ojongerius:
@dtan4 awesome! Would you be able to add a test to [/integration/monitors_test.go](https://github.com/zorkian/go-datadog-api/blob/master/integration/monitors_test.go)?

fprimex:
Hi, as mentioned over in the Terraform Datadog provider issues ( https://github.com/terraform-providers/terraform-provider-datadog/issues/31 ), this PR would resolve an issue importing and converging resources when using Terraform. Is there a timeline for getting this merged? Is there any engineering effort that needs to be contributed to to get this accepted and released? Thanks!

dtan4:
Sorry for late! I added integration testcase to verify this change.

fprimex:
Awesome :) would it be possible to ping us in the TF issue that was linked when a release is made? I'll be keeping an eye out, but we'll probably do a new release immediately after it is available.

ojongerius:
@fprimex have released https://github.com/zorkian/go-datadog-api/releases/tag/v2.6 and gave a heads up in https://github.com/terraform-providers/terraform-provider-datadog/issues/31

PR #138, opened 2017-12-25
Fixing all golint issues.

tariq1890:
@zorkian @ojongerius.  

This PR solves all the golint issues of this repo. Currently, golint is at 57% according to goreportcard.

https://goreportcard.com/report/github.com/zorkian/go-datadog-api#golint

This change will restore it to 100%.

No changes in logic. Only refactors.

### DIFF
--- a/integration/snapshot_test.go
+++ b/integration/snapshot_test.go
@@ -8,11 +8,11 @@ import (
 )
 
 func Jsonify(v interface{}) (string, error) {
-	if b, err := json.MarshalIndent(v, "", " "); err != nil {
+	b, err := json.MarshalIndent(v, "", " ")
+	if err != nil {
 		return "", err
-	} else {
-		return string(b), nil
 	}
+	return string(b), nil
 }
 
 func init() {
@@ -24,11 +24,11 @@ func TestSnapshot(t *testing.T) {
 	end := time.Now().Unix()
 	start := end - 3600
 
-	query_s := "avg:system.mem.used{*}"
-	url, err := client.Snapshot(query_s, time.Unix(start, 0), time.Unix(end, 0), "")
+	q := "avg:system.mem.used{*}"
+	url, err := client.Snapshot(q, time.Unix(start, 0), time.Unix(end, 0), "")
 
 	if err != nil {
-		t.Fatalf("Couldn't create snapshot for query(%s): %s", query_s, err)
+		t.Fatalf("Couldn't create snapshot for query(%s): %s", q, err)
 	}
 
 	fmt.Printf("query snapshot url: %s\n", url)
@@ -42,17 +42,17 @@ func TestSnapshotGeneric(t *testing.T) {
 	// create new graph def
 	graphs := createCustomGraph()
 
-	graph_def, err := Jsonify(graphs[0].Definition)
+	g, err := Jsonify(graphs[0].Definition)
 	if err != nil {
-		t.Fatalf("Couldn't create graph_def: %s", err)
+		t.Fatalf("Couldn't create graphDef: %s", err)
 	}
 
-	options := map[string]string{"graph_def": graph_def}
+	options := map[string]string{"graphDef": g}
 
 	url, err := client.SnapshotGeneric(options, time.Unix(start, 0), time.Unix(end, 0))
 
 	if err != nil {
-		t.Fatalf("Couldn't create snapshot from graph_def(%s): %s", graph_def, err)
+		t.Fatalf("Couldn't create snapshot from graphDef(%s): %s", g, err)
 	}
 
 	fmt.Printf("Graph def snapshot url: %s\n", url)
--- a/alerts.go
+++ b/alerts.go
@@ -15,7 +15,7 @@ import (
 // Alert represents the data of an alert: a query that can fire and send a
 // message to the users.
 type Alert struct {
-	Id           *int    `json:"id,omitempty"`
+	ID           *int    `json:"id,omitempty"`
 	Creator      *int    `json:"creator,omitempty"`
 	Query        *string `json:"query,omitempty"`
 	Name         *string `json:"name,omitempty"`
@@ -34,7 +34,7 @@ type reqAlerts struct {
 // Alert so you can pass that to UpdateAlert later if needed.
 func (client *Client) CreateAlert(alert *Alert) (*Alert, error) {
 	var out Alert
-	if err := client.doJsonRequest("POST", "/v1/alert", alert, &out); err != nil {
+	if err := client.doJSONRequest("POST", "/v1/alert", alert, &out); err != nil {
 		return nil, err
 	}
 	return &out, nil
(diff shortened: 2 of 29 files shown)


yfronto:
From a stylistic perspective, I'm generally a fan. However, this will introduce a large number of breaking changes downstream.

We'll have to take care to tag and pull it in to consumers (like [terraform-provider-datadog](https://github.com/terraform-providers/terraform-provider-datadog)) appopriately.

tariq1890:
Oh yes. That is true. So do we tag some of these consumers and get their opinions on this?  I'd imagine this is no value add for the consumers as this is a cosmetic change.

zorkian:
I'm in favor of the idea, but we can't make breaking changes without doing a major version revision. Is there a way to tell Golint to ignore the kinds of things that would cause breaking changes (such as `Id`->`ID`?)

integration/snapshot_test.go
```diff
@@ -24,11 +24,11 @@ func TestSnapshot(t *testing.T) {
 	end := time.Now().Unix()
 	start := end - 3600
 
-	query_s := "avg:system.mem.used{*}"
-	url, err := client.Snapshot(query_s, time.Unix(start, 0), time.Unix(end, 0), "")
+	queryS := "avg:system.mem.used{*}"
```
ojongerius: Might as well make this `queryString` or `q` as it is very short lived. I prefer `queryString` but Go devs seem to prefer short rather than long if the variable is short lived. See https://github.com/golang/go/wiki/CodeReviewComments#variable-names

integration/snapshot_test.go
```diff
@@ -42,17 +42,17 @@ func TestSnapshotGeneric(t *testing.T) {
 	// create new graph def
 	graphs := createCustomGraph()
 
-	graph_def, err := Jsonify(graphs[0].Definition)
+	graphDef, err := Jsonify(graphs[0].Definition)
```
ojongerius: `graphDefinition` or `g` as its short lived? I prefer `graphDefinition` but Go devs seem to prefer short rather than long if the variable is short lived. See https://github.com/golang/go/wiki/CodeReviewComments#variable-names

ojongerius:
Thanks for this work @tariq1890!

While this is good from a style perspective, it does seem like busy work for consumer when then want to pull in a new version.

I like @zorkian's suggestion to bundle this with other breaking changes. @zorkian did you have anything specific in mind?

integration/snapshot_test.go
tariq1890: I can go with g

integration/snapshot_test.go
tariq1890: I can make it q or perhaps even qs.

tariq1890:
Thank you for all your inputs. I would definitely agree this be piggybacked in a major revision. If there are any changes proposed in the pipeline, I'd be happy to contribute and help.

zorkian:
I don't have any major things in mind. I know `IsAdmin` got deprecated upstream but since it's exported we can't remove it, so that's something pending for a major version change. There are probably some other minor things that we could do but I think overall I haven't had much or heard much in the way of need for major changes. The library is a pretty straightforward shim and does what it needs to do.

I guess if we don't have any pressing needs for a major version update, then we could just bite the bullet and fix the linting issues and roll an optional v3 which people can upgrade to at their leisure. The changes are mostly fairly simple things like naming changes, so it's not a big update anyway. I don't know how complicated the Terraform integration is tho, or how painful those are, so does someone with that knowledge want to make the call on whether a linting-major-revision is worth doing?

yfronto:
On the TF provider side, it's just busy work to update the references anywhere; it's not hard (I mean, the compiler will catch what needs to be changed). It's just a matter of pulling in the latest api version, fixing up the plumbing, and raising a PR (HashiCorp manages the provider repo).

BUT... one thing that could be considered for a major revision is to hide data elements.

We could change the generation code to make data elements private and create getters/setters for all elements instead of just pointers. This opens the door for schema tags on structs to allow for deprecating fields as well as custom demarshaling code for cases like the API returning strings or floats for the same object.

I wasn't originally a fan of using a getter/setter model for a golang library, but lack of strong typing in the API seems to make it necessary more and more.

Thoughts?

tariq1890:
I've rebased this PR and fixed the code as per the review comments by @ojongerius. @yfronto - I like the idea but correct me if I am wrong. If we do make the fields private, this would again affect the many consumers as it is likely that they will use the public members directly instead of the accessors that are available.

yfronto:
@tariq1890 Yes, it would definitely be disruptive, but changing the names on fields and making them private with accessor methods has the same effect downstream. I was just giving some ideas for more things that may make sense to go in the pipeline for major version update. It's easy enough to fix things downstream that we know about (like the Terraform provider), but we can't know all the consumers.

I'm fine merging these to a new version branch for now to start on the next major revision. Not sure that there's a huge list of features, but I think there's probably a good amount of bugfixes that are somewhat blocked on the current version.

ojongerius:
In regards to consumers, I think the Terraform provider is the most active consumer. However I know Datadog has used this library in the past, in house. It looks like they still do publicly in https://github.com/DataDog/kvexpress (`v1`) and https://github.com/DataDog/datadog-agent (`v2`).
I just did a quick GH code search and did notice there are a lot more consumers than I imagined! I see Mozilla, Concourse, Zendesk, InvisionApp, Cloudfoundry using it in projects, and see government usage too. 246 results for "github.com/zorkian/go-datadog-api" and 71 for "gopkg.in/zorkian/go-datadog-api" this is just a rough query using the web code search Github provides. Those results have not been deduped and include inactive projects. If someone feels like digging a little deeper, feel free 😄 

I see merit in @yfronto's idea. @yfronto would you be willing to take on the work and have time for it in the near future?

yfronto:
Wow, I had no idea there were so many consumers. @ojongerius Sure, I'm happy to take that. I should be able to get it done reasonably quickly, maybe in the next few weeks at the latest. Probably sooner, depending on timing of other projects.

ojongerius:
@yfronto cool. In that case merging this PR into a new version branch and implementing your changes there seems like the way forward. @tariq1890 and @zorkian are you guys onboard with that?

zorkian:
Wow. I also had no idea this was used so widely. That's cool.

I'm very on board with merging to a v3 branch and making the breaking changes there, but would like to make sure we cap the investment into ongoing v2 work so that we don't have to maintain two versions forever. (We should definitely apply bugfixes to v2, but make everybody do feature stuff in v3 once we get it going.)

I'm also open to the idea of hiding the data fields and just having accessors so we can change behaviors/add deprecation warnings. One thing it does complicate is the constructors though, because today you can just instantiate the objects and that's actually fairly nice. Do you have any thoughts on how you might approach that without making it way too painful?

yfronto:
**tl;dr**: Two approaches: make everything private and generate accessors for everything, which is easy but means using constructors, or leave things public and generate marshal/demarshal code that consumes struct field tags to make decisions, which is a tad harder but leaves consumers writing more idiomatic code.

**Full discussion**

We could generate constructors along with accessors, but we would lose the ability to initialize with a static struct definition. I don't know that there's a clean workaround for that. 

I was thinking of generating something like the following for each type, which follows what I've seen in other Go libs that use this pattern:

```golang
type Foo struct {
  bar *string
}

func (f *Foo) Bar() (string, bool) {
  # Get value of bar and true or "" and false
}

func (f *Foo) SetBar(s string) {
  # Sets bar
}

func NewFoo(s string) *Foo {
  return &Foo{s}
}
```

I don't mind this pattern and wish variatic returns were a thing, but it leaves consumers calling constructors and writing code that looks a bit less like go.

Alternatively, we could leave type fields public and change the generator code to generate the marshal/unmarshal receiver funcs for each type. The marshal code could take certain actions or notify based on struct field tags that we come up with. Similarly, the demarshal code could handle weird typing quirks that arise more elegantly. 

It means that any logic about the type only takes place at the marshal/unmarshal call site and not when getting/setting member fields. This is how some serialization libraries (like protos) already function (ish), and I'd be fine with this, although it makes the generation code a tad more complicated.

yfronto:
I created a `next` branch for this PR. I'll work from this branch to make some changes we discussed.

PR #140, opened 2018-02-11
Allow monitor tags to be removed on update

MattHodge:
# Issue

* You create a monitor that has tags
* You remove the tags from the monitor and call `UpdateMonitor`
* The monitor's tags are not updated as an empty `tags` isn't passed to the DataDog API due to it being omitted in the `Monitor` struct when marshalling to JSON

# Fix

* Changed the `Monitor` struct to not omit empty `tags` when marshalling to JSON

### DIFF
--- a/integration/monitors_test.go
+++ b/integration/monitors_test.go
@@ -50,6 +50,25 @@ func TestMonitorUpdate(t *testing.T) {
 
 }
 
+func TestMonitorUpdateRemovingTags(t *testing.T) {
+
+	monitor := createTestMonitorWithTags(t)
+	defer cleanUpMonitor(t, *monitor.Id)
+
+	monitor.Tags = make([]string, 0)
+	if err := client.UpdateMonitor(monitor); err != nil {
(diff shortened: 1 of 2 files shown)


MattHodge:
# Issue

* You create a monitor that has tags
* You remove the tags from the monitor and call `UpdateMonitor`
* The monitor's tags are not updated as an empty `tags` isn't passed to the DataDog API due to it being omitted in the `Monitor` struct when marshalling to JSON

# Fix

* Changed the `Monitor` struct to not omit empty `tags` when marshalling to JSON

yfronto:
LGTM. Thanks for fixing this!

@zorkian @ojongerius @nyanshak : For future tags, I'd like to switch to semantic versioning, so I'm going to tag this as v2.8.1. Let me know if you have any objections to this.

zorkian:
What do you mean 'switch to'? Oh, because we never had a patch number on the version? I guess I never realized that SemVer requires X.Y.Z form!

Anyway, I'm totally in support of Semantic Versioning. Let's do it.

PR #141, opened 2018-02-14
Add Unit extraction to the query get api

tjoyal:
:wave: I'm pretty new to go so tell me to "go" away if this crazy talk 😸 

The query response would look like this:
```json
{
  "status": "ok",
  "res_type": "time_series",
  "series": [
    {
      "metric": "...",
      "unit": [
        {
          "family": "time",
          "scale_factor": 0.001,
          "name": "millisecond",
          "short_name": "ms",
          "plural": "milliseconds",
          "id": 10
        },
        null
      ],
      "pointlist": [...],
    }
  ]
}
```

Where the `unit` entry can either be `null` or an array of 2 "unit", where each of these units could be `nil`.

Does it make any sense?

### DIFF
--- a/series.go
+++ b/series.go
@@ -31,6 +31,7 @@ type Metric struct {
 // Unit represents a unit definition that we might receive when query for timeseries data.
 // A metric is characterized by 2 units as: x per y
 // One or both could be missing
+type UnitPair *[2]*Unit
 type Unit struct {
 	Family string       `json:"family"`
 	ScaleFactor float32 `json:"scale_factor"`
@@ -52,7 +53,7 @@ type Series struct {
 	Length      *int        `json:"length,omitempty"`
 	Scope       *string     `json:"scope,omitempty"`
 	Expression  *string     `json:"expression,omitempty"`
-	Units       *[2]*Unit   `json:"unit"`
+	Units       UnitPair    `json:"unit"`
 }
 
 // reqPostSeries from /api/v1/series
--- a/series.go
+++ b/series.go
@@ -29,9 +29,6 @@ type Metric struct {
 }
 
 // Unit represents a unit definition that we might receive when query for timeseries data.
-// A metric is characterized by 2 units as: x per y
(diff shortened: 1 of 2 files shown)


series.go
```diff
@@ -40,6 +52,7 @@ type Series struct {
 	Length      *int        `json:"length,omitempty"`
 	Scope       *string     `json:"scope,omitempty"`
 	Expression  *string     `json:"expression,omitempty"`
+	Units       *[2]*Unit   `json:"unit"`
```
yfronto: This isn't incredibly readable. I'd rather it be declare a UnitPair type that's an alias for a slice of Unit pointers and have this be a pointer to that type.

series.go
```diff
@@ -31,6 +31,7 @@ type Metric struct {
 // Unit represents a unit definition that we might receive when query for timeseries data.
```
yfronto: This comment describes the Unit struct but it's above UnitPair

series.go
```diff
@@ -31,6 +31,7 @@ type Metric struct {
...
 // A metric is characterized by 2 units as: x per y
 // One or both could be missing
+type UnitPair *[2]*Unit
```
yfronto: The DD API isn't ideal for Go, so I'm not sure the best approach... but....

Instead of having an array of size 2 here, use a slice:
```go
type UnitPair []*Unit
```
And define a receiver method on Unit to manage it:
```go
func (u *Unit) SetUnit(unit, per_unit *Unit) {
  // Set u.Units appropriately
}
```

That's probably the best way to keep it understandable, given the DD API.

tjoyal: Awesome, that was the pattern I was looking for. I'll try to get around to do it sometimes soon.

I'm still in the process of figuring out how to properly test things and hook up fixtures to properly test this PR.

yfronto:
Hi @tjoyal, any update on this?

yfronto:
LGTM. Thanks @tjoyal!

PR #142, opened 2018-02-14
Allow programatic set of base url

tjoyal:
`baseUrl` to be changeable through `Client.SetBaseUrl`

Since app_key and api_key are set programatically, I don't see a reason why we wouldn't allow the same for the domain. This will allow us to handle the base url the same as we handle the keys and not in a different way.

Made the in a way where it should not break anything for existing users (except if you were dynamically changing your env variable after the client was constructed). 

Could have been simply added to the constructor but I'm not sure how to handle breaking changes in the go community or if it really made sense as most people would use the default value.

Thoughts?

### DIFF
--- a/client.go
+++ b/client.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"net/http"
 	"time"
 )
@@ -20,7 +21,7 @@ import (
 // Client is the object that handles talking to the Datadog API. This maintains
 // state information for a particular application connection.
 type Client struct {
-	apiKey, appKey string
+	apiKey, appKey, baseUrl string
 
 	//The Http Client that is used to make requests
 	HttpClient   *http.Client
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -36,20 +37,31 @@ type valid struct {
...
 		appKey:       appKey,
+		baseUrl:      baseUrl,
 		HttpClient:   http.DefaultClient,
 		RetryTimeout: time.Duration(60 * time.Second),
 	}
 }
 
 // SetKeys changes the value of apiKey and appKey.
-func (c *Client) SetKeys(apiKey, appKey string) {
+func (c *Client) SetKeys(apiKey string, appKey string) {
```
yfronto: This extra type declaration is unnecessary.

request_test.go
```diff
@@ -12,6 +12,7 @@ func TestUriForApi(t *testing.T) {
 	c := Client{
 		apiKey:       "sample_api_key",
 		appKey:       "sample_app_key",
+		baseUrl:      "https://app.datadoghq.com",
```
yfronto: We'll also want a test that doesn't include a baseUrl.

yfronto:
Thanks for doing this. I can see it being useful if DD adds subdomains, which I suspect they might do eventually, since their IDs are not globally unique and some companies have multiple team accounts.

PR #144, opened 2018-03-10
Add integrations/pagerduty API Support

rlhh:
Datadog has recently introduced APIs to support CRUD operations for integrations(https://docs.datadoghq.com/api/?lang=bash#integrations). 

This PR adds support for CRUD operations on the Pagerduty Integration https://docs.datadoghq.com/api/?lang=bash#pagerduty. 

Things to note:
1) The request and response body does not match 1:1. For the request, the "service" object has "service_name" and "service_key" but for the response, we get "service" and "key". 
2) The integration tests  only compare the service names because 
- The service key is returned as ****
- The schedules are verified so we cannot put in fake data during creation
- The API key is returned as ****

### DIFF
--- a/integrations.go
+++ b/integrations.go
@@ -8,45 +8,45 @@
 
 package datadog
 
-type reqServicePD struct {
-	ServiceName string `json:"service"`
-	ServiceKey  string `json:"key"`
+type servicePD struct {
+	ServiceName *string `json:"service"`
+	ServiceKey  *string `json:"key"`
 }
 
-type reqIntegrationPD struct {
-	Services  []reqServicePD `json:"services"`
-	Subdomain string         `json:"subdomain"`
-	Schedules []string       `json:"schedules"`
-	APIToken  string         `json:"api_token"`
+type integrationPD struct {
+	Services  []servicePD `json:"services"`
+	Subdomain *string     `json:"subdomain"`
+	Schedules []string    `json:"schedules"`
+	APIToken  *string     `json:"api_token"`
 }
 
-type ServicePD struct {
-	ServiceName string `json:"service_name"`
-	ServiceKey  string `json:"service_key"`
+type ServicePDRequest struct {
+	ServiceName *string `json:"service_name"`
+	ServiceKey  *string `json:"service_key"`
 }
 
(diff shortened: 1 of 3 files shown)


integrations.go
```diff
@@ -0,0 +1,60 @@
+/*
+ * Datadog API for Go
+ *
+ * Please see the included LICENSE file for licensing information.
+ *
+ * Copyright 2013 by authors and contributors.
```
ojongerius: Could you please update the year to 2018?

integrations.go
```diff
@@ -0,0 +1,60 @@
...
+	ServiceKey  *string `json:"key"`
+}
+
+type integrationPD struct {
+	Services  []servicePD `json:"services"`
+	Subdomain *string     `json:"subdomain"`
+	Schedules []string    `json:"schedules"`
+	APIToken  *string     `json:"api_token"`
+}
+
```
ojongerius: Could you please add comments? See https://blog.golang.org/godoc-documenting-go-code

integrations.go
```diff
@@ -0,0 +1,60 @@
...
+func (client *Client) GetIntegrationPD() (*integrationPD, error) {
+	var out integrationPD
+	if err := client.doJsonRequest("GET", "/v1/integration/pagerduty", nil, &out); err != nil {
+		return nil, err
+	}
+
+	return &out, nil
+}
+
+// DeleteIntegrationPD remove the PD Integration from the system.
```
ojongerius: "removes"

ojongerius:
Thanks for this!

> the request and response body does not match 1:1. For the request, the "service" object has "service_name" and "service_key" but for the response, we get "service" and "key".

Would you be willing to raise an issue at Datadog, possibly linking this PR in, that looks like a bug to me.

rlhh:
Thanks for the review! 

I'm not sure what's the best place to raise this issue so I've raised it at datadog/documentation. It should at minimum be documented on Datadog itself.

rlhh:
@ojongerius @yfronto What would be the next steps for this PR?

ojongerius:
Thanks @rlhh! 🙇

PR #147, opened 2018-03-26
Use strings in Yaxis struct to workaround error

robinbowes:
"json: cannot unmarshal string into Go struct field Yaxis.min of type
float64"

See: https://github.com/zorkian/go-datadog-api/issues/103

I have not tested this at all - my go-fu is weak and I don't know how to vendor this locally and build the terraform datadog provider to use it to test it.

If anyone can point me in the right direction, I'll do more testing...

### DIFF
--- a/dashboards.go
+++ b/dashboards.go
@@ -52,9 +52,9 @@ type GraphEvent struct {
 }
 
 type Yaxis struct {
-	Min   *float64 `json:"min,omitempty"`
-	Max   *float64 `json:"max,omitempty"`
-	Scale *string  `json:"scale,omitempty"`
+	Min   *string `json:"min,omitempty"`
+	Max   *string `json:"max,omitempty"`
+	Scale *string `json:"scale,omitempty"`
 }
 
 type Style struct {
--- a/datadog-accessors.go
+++ b/datadog-accessors.go
@@ -12786,18 +12786,18 @@ func (w *Widget) SetToplistWidget(v ToplistWidget) {
 }
 
 // GetMax returns the Max field if non-nil, zero value otherwise.
-func (y *Yaxis) GetMax() float64 {
+func (y *Yaxis) GetMax() string {
 	if y == nil || y.Max == nil {
-		return 0
+		return ""
 	}
 	return *y.Max
 }
 
 // GetOkMax returns a tuple with the Max field if it's non-nil, zero value otherwise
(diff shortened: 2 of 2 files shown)


yfronto:
Hi @robinbowes, thanks for working on this! I've meant to fix this for a while but haven't had the chance. 

The problem with changing the types of these structs is that it's a breaking change downstream, so not much of a bugfix. I think the right approach is probably to add a boolean "auto" to the Yaxis type and define custom marshalJSON/unmarshalJSON functions. So, the struct becomes:

```go
type Yaxis struct {
  Min   *float64
  AutoMin *bool
  Max   *float64
  AutoMax *bool
  ...
 }
```

You can define a custom unmarshal on the Yaxis type that wraps an alias type but overrides the particular fields during the unmarshal:

```go
func (y *Yaxis) UnmarshalJSON(data []byte) error {
  type Alias Yaxis
  wrapper := &struct {
    Min *string `json:"min"`
    Max *string `json:"max"`
    *Alias
  }{
    Alias: (*Alias)(y),
  }
  if err := json.Unmarshal(data, wrapper); err != nil {
     return err
  }
  // Now look at aux.Min, which is a *string, and see if it's "auto" if so...
  auto := true
  i.Auto = &auto
  i.Min = nil
  // Otherwise, explicitly set it to nil and set i.Min
  i.Auto = nil
  f, err := strconv.ParseFloat(aux.Min, 64)
  // Check err
  i.Min = &f  
  // Same with Max
}
```

With the Marshal, do a similar pattern, checking if the auto field is true, otherwise marshaling the float as is. Since the default for a *bool will be false, existing code shouldn't break.

Does that make sense?

robinbowes:
Hey @yfronto , thanks for the reply and sorry I've not got back to you.

What you say sort of makes sense, but I'm not sure I know enough go to knock up a solution.

I'll see if I can make time over the next week or so.

Thanks again,

R.

robinbowes:
I'm still interested in this, but it's down my priority list and so unlikely to get any of my time for the forseeable. Sorry.

R.

jamesmharvey:
@robinbowes I was talking to @atsang about this, and it sounds like he's interested in trying to make some progress on it. Does that work for you?

yfronto:
Fixed in #167

robinbowes:
Thanks for picking this up @anthony-tsang

PR #150, opened 2018-04-07
Add retry logic to client's Validate method

GrantSheehan:
In order to get around Datadog's API throwing 503 errors with some
regularity, adds retry logic for retryable errors.

Background:
I use the Datadog Terraform provider and it's written in a way that most of the failed requests will be retried without much issue, but refreshing the credentials on the provider itself will run the `Validate()` method, which is currently not able to be retried, which ends up failing quite a few of our CI/CD jobs. This seemed like the most pragmatic way to resolve some of those errors (as Datadog's support has been somewhat less than helpful in resolving the cause of the 503's themselves, *cough*). If you don't like the way it's implemented here, just let me know what changes you'd like made I'll update it.

Thanks!


client.go
```diff
@@ -85,20 +91,24 @@ func (client *Client) Validate() (bool, error) {
 	}
 
 	var resp *http.Response
-	resp, err = client.HttpClient.Do(req)
-	if err != nil {
-		return false, err
-	}
 
-	defer resp.Body.Close()
```
ojongerius: Why not defer closing `resp.Body`?

GrantSheehan: I was getting errors in unmarshaling JSON during testing, I think it was because I had it in the wrong scope. I've added it back.

GrantSheehan:
I used the expontential backoff retrier package being used in `request.go` and mirrored more or less what it was doing.

ojongerius:
Fair enough, that is indeed what we used before. Just stumbled upon go-retryablehttp and thought it looked neat.

GrantSheehan:
Yeah, for sure, looks like it could make the code significantly easier to reason about. I can take a crack at migrating to it after this is merged, if you'd like.

ojongerius:
Ok this is entertaining; I wrote the initial validate() method, and should have used `doRequestWithRetries` instead of doing the request manually. Thank you for wanting for fix the bug I created! 😊 

Would you mind refactoring this to use `doRequestWithRetries` instead of copying the retry logic, or do you see a reason why that would not be desirable?

GrantSheehan:
I updated my branch to use the doRequestWithRetries method instead.

GrantSheehan:
@ojongerius Is there any additional testing you'd like me to do here? Am I missing anything?

GrantSheehan:
Sorry to be a pain, this has been a thorn in our side for a while now. Having finally nailed down a root cause and a fix, I'll admit to being a bit eager to get a fix merged.

Glad you're feeling better!

client.go
```diff
@@ -69,43 +67,28 @@ func (c *Client) GetBaseUrl() string {
...
-
-	// Only care about 200 OK or 403 which we'll unmarshal into struct valid. Everything else is of no interest to us.
-	if resp.StatusCode != 200 && resp.StatusCode != 403 {
-		body, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			return false, err
-		}
-		return false, fmt.Errorf("API error %s: %s", resp.Status, body)
+	if body, err := ioutil.ReadAll(resp.Body); err != nil {
```
ojongerius: I don't think this will work, as we'll need `body` later on.

client.go
```diff
@@ -69,43 +67,28 @@ func (c *Client) GetBaseUrl() string {
...
+		return false, err
 	}
 
-	body, err := ioutil.ReadAll(resp.Body)
-	err = json.Unmarshal(body, &out)
-	if err != nil {
+	if err = json.Unmarshal(body, &out); err != nil {
```
ojongerius: 👍

client.go
```diff
@@ -69,43 +67,28 @@ func (c *Client) GetBaseUrl() string {
...
 		return false, err
 	}
 
+	defer resp.Body.Close()
```
ojongerius: Please move the defer up to just after checking if `doRequestWithRetries` return an error, I think below line 82.

client.go
GrantSheehan: Forgive my golang ignorance, how does doing it in the `if` change the logic this way?

GrantSheehan:
Yep, no problem. I'm still trying to grok some of golang's idioms/syntax, be as nit-picky as you want to be.

client.go
```diff
@@ -69,43 +67,31 @@ func (c *Client) GetBaseUrl() string {
...
 	}
-	req, err := http.NewRequest("GET", uri, bodyreader)
 
-	if err != nil {
-		return false, err
-	}
-	if bodyreader != nil {
-		req.Header.Add("Content-Type", "application/json")
-	}
+	req, err := http.NewRequest("GET", uri, nil)
```
ojongerius: Still needs checking if we got an error here.

@GrantSheehan btw, you were absolutely right about inline error checking. I stand corrected!

ojongerius:
Nice work @GrantSheehan!

GrantSheehan:
I've been testing by vendoring and installing my forked repo/branch when compiling the datadog Terraform provider and have had no issues. If there's anything you'd like me to do in addition to that, let me know.

ojongerius:
I've cut release [2.8.4](https://github.com/zorkian/go-datadog-api/releases/tag/v2.8.4).

GrantSheehan:
Thank you!

PR #157, opened 2018-05-30
Adding Integration Tests in TravisCI

MattHodge:
* Adds the integration tests to TravisCI (4 currently fail).
* Hopefully, this helps with #39.

You just need to add `DATADOG_API_KEY` and `DATADOG_APP_KEY` to the [TravisCI environment variables](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings).

PR #158, opened 2018-06-15
Fix Invalid screen board JSON structure

rodrigorznd:
Fix #32

PR #161, opened 2018-06-22
Fixing integration tests (2/2)

### LINKED ISSUE #74: Schedule integration test runs
This should be valuable enough, we could use https://docs.travis-ci.com/user/cron-jobs . 

- [x] arrange sandbox credentials at Datadog.
- [ ] setup a schedule for these builds.

yfronto:
This fixes the remaining integration test failures following #157. The user creation fails to delete a user. This is because DD doesn't delete users; it disables them. When the test account is created,
it's already been created and disabled. To workaround the lack of deletion support (and to avoid creating a user every run), the test explicitly enables the user after creation. The other failure was for
the signature of the payload in the PagerDuty integration, which has been fixed. This commit should cut a new release, since it looks like this has been a bug for a while.

Closes #74

* User create/delete tests now enable user after creation
* PagerDuty integration updated

### DIFF
--- a/integration/client_test.go
+++ b/integration/client_test.go
@@ -8,10 +8,6 @@ import (
 	"github.com/zorkian/go-datadog-api"
 )
 
-func init() {
-	client = initTest()
-}
-
 func TestInvalidAuth(t *testing.T) {
 	// Override the correct credentials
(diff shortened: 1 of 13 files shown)


yfronto:
@nyanshak This should unblock running integration tests for #158 once it's merged.

PR #163, opened 2018-08-18
Add monitor state to monitor results

Seanstoppable:
This allows one to use the api to trigger actions based on a monitor in
'Alert' status, as an example

### DIFF
--- a/monitors.go
+++ b/monitors.go
@@ -60,15 +60,15 @@ type Options struct {
 // Monitor allows watching a metric or check that you care about,
 // notifying your team when some defined threshold is exceeded
 type Monitor struct {
-	Creator *Creator `json:"creator,omitempty"`
-	Id      *int     `json:"id,omitempty"`
-	Type    *string  `json:"type,omitempty"`
-	Query   *string  `json:"query,omitempty"`
-	Name    *string  `json:"name,omitempty"`
-	Message *string  `json:"message,omitempty"`
-	State   *string  `json:"overall_state,omitempty"`
-	Tags    []string `json:"tags"`
-	Options *Options `json:"options,omitempty"`
+	Creator      *Creator `json:"creator,omitempty"`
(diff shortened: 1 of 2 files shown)


monitors.go
```diff
@@ -66,6 +66,7 @@ type Monitor struct {
 	Query   *string  `json:"query,omitempty"`
 	Name    *string  `json:"name,omitempty"`
 	Message *string  `json:"message,omitempty"`
+	State   *string  `json:"overall_state,omitempty"`
```
ojongerius: When checking the DD API docs at https://docs.datadoghq.com/api/?lang=bash#monitors I noticed both "state" and "overal_state" exist, and have different meanings. Unmarshalling "overall_state" into "State" would be confusing.

ojongerius:
Thanks for taking the time to create this PR @Seanstoppable!

When reading up on the API doco at https://docs.datadoghq.com/api/?lang=bash#monitors I noticed there are quite a few changes possible in regards to states:

## Input
* `group_states`: could be "all" or "alert,warn" used as a an argument to getting monitor details

## Output
* `state`: empty or contains an array of `groups`, consisting of `hosts`
* `overall_state`: `*string`, example "Alert" -as seen in example of "resolve monitor"
* `overall_state_modified`:  `*string`: "2017-08-11T12:21:05.949182+00:00" -as seen in example of "resolve monitor"

I'be be happy to merge this afer you address my inline comment. However it would be wonderful if you had the time to implement `state, `overall_state` and `overal_state_modified`.

monitors.go
Seanstoppable: Completely missed regular state, thanks! I've updated this to be inline with the json.

Seanstoppable:
I'm happy to look at further expansion in my spare time. `OverallState` is just something I am trying to access for another project.

ojongerius:
@Seanstoppable great! I've raised #164 feel free to call dibs on it.

Issue #32, opened 2016-01-06
Invalid screen board JSON structure


seiffert:
I'm experiencing several issues with the screen boards API which was implemented in #13. From what I can tell, the screen board JSON documents that are being sent to DataDog are invalid. 
IMO the tests pass because the Datadog API is more or less a simple key-value store that saves everything you throw at it. However, I can't make it render screenboards created with the new data structure.

Compare those two JSON representations of the same widget:

The one that works (generated with the 'old' screen board code):

```
    {
      "title_size": 16,
      "title": true,
      "title_align": "left",
      "title_text": "# Containers",
      "height": 13,
      "tile_def": {
        "viz": "timeseries",
        "requests": [
          {
            "q": "...",
            "style": {
              "palette": "blue"
            },
            "type": "bars"
          }
        ],
        "events": [
          {
            "q": "tags:..."
          }
        ],
        "markers": []
      },
      "width": 36,
      "timeframe": "1d",
      "y": 5,
      "x": 111,
      "type": "timeseries",
      "legend": false
    }
```

vs. the one generated with the current go-datadog-api and the new Widget data structure. This one does not work:

```
    {
      "alert_graph": {},
      "iframe": {},
      "graph": {
        "tile_def": {}
      },
      "image": {
        "title_size": {
          "Auto": false,
          "Size": 0
        }
      },
      "alert_value": {},
      "change": {
        "tile_def": {}
      },
      "query_value": {
        "text_size": {
          "Auto": false,
          "Size": 0
        },
        "title_size": {
          "Auto": false,
          "Size": 0
        }
      },
      "hostmap": {
        "tile_def": {}
      },
      "free_text": {},
      "timeseries": {
        "title_size": {
          "Auto": false,
          "Size": 16
        },
        "title": true,
        "title_align": "left",
        "title_text": "# Containers",
        "height": 13,
        "tile_def": {
          "viz": "timeseries",
          "requests": [
            {
              "q": "...",
              "style": {
                "palette": "blue"
              },
              "type": "bars"
            }
          ],
          "events": [
            {
              "q": "tags:..."
            }
          ]
        },
        "width": 36,
        "timeframe": "1d",
        "y": 5,
        "x": 111,
        "type": "timeseries"
      },
      "check_status": {},
      "event_stream": {
        "title_size": {
          "Auto": false,
          "Size": 0
        }
      },
      "event_timeline": {},
      "frame": {},
      "toplist": {
        "tile_def": {},
[...]

PR #168, opened 2018-09-04
Fix Screenboard support

oxlay:
Datadog engineer here 👋 

As explained in #32, the support for Datadog screenboards was broken. This PR adresses this issue.

# Things changed

* Some struct fields have been removed to match removed attributes in the API, and some struct fields have been added to add support for new widgets. **All 19 screenboard widgets have been tested sucessfully.**
* The multiple screenboard widgets have been merged into a single one, to avoid redundancy and to match more closely the actual API (which accepts pretty much any JSON)
* Integration tests for screenboard widgets have been removed. Indeed, they would not really test anything since the API accepts arbitrary JSON.


screen_widgets.go
```diff
@@ -1,287 +1,201 @@
...
-	Y          *int    `json:"x,omitempty"`
-	Tags       *string `json:"tags,omitempty"`
-	Timeframe  *string `json:"timeframe,omitempty"`
-	TextSize   *string `json:"text_size,omitempty"`
-	Type       *string `json:"type,omitempty"`
-	Check      *string `json:"check,omitempty"`
-	Group      *string `json:"group,omitempty"`
-	Grouping   *string `json:"grouping,omitempty"`
+type Time struct {
+	LiveSpan *string `mapstructure:"live_span" json:"live_span,omitempty"`
```
nyanshak: There's no reference to the mapstructure package in this repo. What's the intent of this tag? Will it be consumed downstream?

cc @yfronto

oxlay: My mistake, didn't add this on purpose. Leftover from #158. Nice catch!

yfronto:
Thanks @oxlay! lgtm

PR #169, opened 2018-09-07
Fix type issue for threshold in screenboard rule

oxlay:
Because of a type issue, we could not submit thresholds such as `99.9` for rules in the uptime widget.

This PR fixes that.

### DIFF
--- a/screen_widgets.go
+++ b/screen_widgets.go
@@ -191,7 +191,7 @@ type Params struct {
 }
 
 type Rule struct {
-	Threshold *int    `json:"threshold,omitempty"`
+	Threshold *string `json:"threshold,omitempty"`
 	Timeframe *string `json:"timeframe,omitempty"`
 	Color     *string `json:"color,omitempty"`
 }
--- a/screen_widgets.go
+++ b/screen_widgets.go
@@ -191,9 +191,9 @@ type Params struct {
 }
 
 type Rule struct {
-	Threshold *string `json:"threshold,omitempty"`
-	Timeframe *string `json:"timeframe,omitempty"`
-	Color     *string `json:"color,omitempty"`
+	Threshold *float64 `json:"threshold,omitempty"`
+	Timeframe *string  `json:"timeframe,omitempty"`
+	Color     *string  `json:"color,omitempty"`
 }
 
 type ScreenboardMonitor struct {
--- a/screen_widgets.go
+++ b/screen_widgets.go
@@ -1,5 +1,7 @@
 package datadog
 
+import "encoding/json"
+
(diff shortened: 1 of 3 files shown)


screen_widgets.go
```diff
@@ -191,7 +191,7 @@ type Params struct {
 }
 
 type Rule struct {
-	Threshold *int    `json:"threshold,omitempty"`
```
nyanshak: Is there a reason this needs to be a string rather than a `*json.Number`? I'd prefer to be more specific about types rather than use stringly-typed variables.

nyanshak:
I'm having trouble finding any documentation about an "uptime" screen widget, so any testing I've been doing to validate this has been based on the code in the integration tests. Do you have any doc links you could share to make this a bit easier?

screen_widgets.go
oxlay: Good point. I've changed it to a `*float64` instead of a `*json.Number` if that's okay, so that people don't have to import `json` to use it.

oxlay:
@nyanshak Actually the `uptime` widget is purposefully undocumented and still a WIP at this point. Sorry about that, missed it in the first PR. I added a warning in the code telling people not to rely too much on it.

nyanshak:
The reason for using `json.Number`, at least across many of Datadog's API endpoints, is that it's common for the API to accept (and send) multiple types. 

I just tested against the screenboard API to make a screenboard with uptime widget with thresholds:
```
    {
      ... truncated ...
      "rules": {
        "1": {
          "threshold": "98", // Hmm.. this can be a string
          "color": "red",
          "timeframe": "7 days"
        },
        "0": {
          "threshold": "95.1", // Hmm, this can also be a string
          "color": "green",
          "timeframe": "Month-to-date"
        }
      },
      "type": "uptime"
    }
```

I'm able to send thresholds as strings (and Datadog subsequently sends them back as strings). Using float64 as the type results in errors somewhat like this:

```
json: cannot unmarshal string into Go struct field Rule.Threshold of type float64
```

Looking back, I believe this is originally the reason we started using `json.Number` rather than `float64`, as this case is handled properly there.

screen_widgets.go
```diff
@@ -191,9 +191,9 @@ type Params struct {
 }
 
 type Rule struct {
-	Threshold *string `json:"threshold,omitempty"`
```
nyanshak: Please use `json.Number` instead of `string` or `float64` here. I've put a long-form comment explaining why on the PR itself.

oxlay: Oh I get it. Works for me. PR updated.

PR #178, opened 2018-10-10
Fix hostmap style setting.

AlexJF:
Hi again :)

I've noticed some other issues with the hostmap widget API. Namely, conflicts between `style` and `Style` keys and missing `fillMin` and `fillMax` options.

When adding a hostmap widget to a dashboard using the current go-datadog-api with a custom palette, the resulting JSON in Datadog looks like the following:

```
{
  "Style": {
    "palette": "green_to_orange"
  },
  (...)
  "style": {
    "fillMin": null,
    "fillMax": null
  }
}
```

PR #179, opened 2018-10-10
Change GraphDefinition.Precision type

nyanshak:
* Datadog's API will accept / return the following types for precision
(query value widget on timeboards / screenboards): int (e.g., 4), string
with integer inside (e.g., "4"), or the specific string "*" (meaning
full precision). This commit adds support for the integer case.


ojongerius:
@nyanshak are you good to merge and cut a release? If so that release would include this changes and #178

Issue #164, opened 2018-08-21
State support


ojongerius:
Follow up from #163 

> When reading up on the API doco at https://docs.datadoghq.com/api/?lang=bash#monitors I noticed there are quite a few changes possible in regards to states:

# Input
- [ ] group_states: could be "all" or "alert,warn" used as a an argument to getting monitor details

# Output
- [ ] state: empty or contains an array of groups, consisting of hosts
- [ ] overall_state: *string, example "Alert" -as seen in example of "resolve monitor"
- [ ] overall_state_modified: *string: "2017-08-11T12:21:05.949182+00:00" -as seen in example of "resolve monitor"

PR #180, opened 2018-10-13
Add full param options for monitors

Seanstoppable:
Add a more generic function that will let people mix and match
Have existing functions wrap this better function
Resolves the rest of state support in #164

### DIFF
--- a/monitors.go
+++ b/monitors.go
@@ -135,7 +135,7 @@ func (client *Client) GetMonitor(id int) (*Monitor, error) {
 
 // GetMonitorsByName retrieves monitors by name
 func (client *Client) GetMonitorsByName(name string) ([]Monitor, error) {
-	return client.GetMonitorsWithOptions(MonitorQueryOpts{Name: name})
+	return client.GetMonitorsWithOptions(MonitorQueryOpts{Name: &name})
 }
 
 // GetMonitorsByTags retrieves monitors by a slice of tags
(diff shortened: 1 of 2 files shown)


Seanstoppable:
Great, thanks @ojongerius

monitors.go
```diff
@@ -171,11 +151,59 @@ func (client *Client) DeleteMonitor(id int) error {
...
+		value := fmt.Sprintf("group_states=%v", strings.Join(opts.GroupStates, ","))
+		query = append(query, value)
+	}
+
+	if len(opts.MonitorTags) > 0 {
+		value := fmt.Sprintf("monitor_tags=%v", strings.Join(opts.MonitorTags, ","))
+		query = append(query, value)
+	}
+
+	if opts.WithDowntimes != nil {
```
ojongerius: Up to you but you could use the generated GetWithDownTimesOk like so:

```go
if v, ok := opts.GetWithDowntimesOk; ok {
    query = append(query, fmt.Sprintf("name=%t", strings.Join(v, ",")))
}
```

monitors.go
```diff
@@ -171,11 +151,59 @@ func (client *Client) DeleteMonitor(id int) error {
...
+		value := fmt.Sprintf("with_downtimes=%t", *opts.WithDowntimes)
+		query = append(query, value)
+	}
+
+	if opts.Name != "" {
```
ojongerius: Works at the moment but needs to be changed either manually test if not nil if you refactor Name to be a pointer. Alternatively you can use the autogenereated getters and setters like so:

```go
if v, ok := opts.GetNameOk; ok {
    query = append(query, fmt.Sprintf("name=%v", strings.Join(v, ",")))
}
```

Seanstoppable:
@ojongerius apologies for the delay. Updated with feedback, thanks!

ojongerius:
@Seanstoppable sorry I've been swamped with work, could you please resolve conflicts so I can merge this?

Seanstoppable:
Closing in favor of #199

Issue #185, opened 2018-10-16
List all dashboards does not include all fields


mnaboka:
[DashboardLite](https://github.com/zorkian/go-datadog-api/blob/master/dashboards.go#L164-L169) structure has very limited number of fields.

I need just one additional field which is `modified`, but ideally i think we need all of them.
Was there a reason not all fields are exported?

I can also open a PR to address this, but i want to hear your feedback first.

Thanks

ojongerius:
Hi @mnaboka! I'm not sure what the reason for this is, the current version of the `DashboardLite` struct is very similar to the initial commit by @zorkian.

Looks like `Dashboard` could do with an update too, `width`, `height`, `read_only` are not included.

mnaboka:
@ojongerius i need to detect if/when dashboard was modified.

ojongerius:
@mnaboka that makes sense, be great if you could raise a PR for that! 🙇

mnaboka:
@ojongerius i opened a [PR 186](https://github.com/zorkian/go-datadog-api/pull/186), would be nice if you'd have a look.

Also it seems that CI is failing, but the tests pass, i am not sure if CI was broken on master or my change did something to it?

Thanks

ojongerius:
Went in with https://github.com/zorkian/go-datadog-api/releases/tag/v2.17.0 

Thanks for your contribution @mnaboka!

PR #186, opened 2018-10-17
Add missing dashboard fields

mnaboka:
Issue #185

If applied this Pull request will add some of the missing fields to `DashboardLite` and adds a generated `datadog-accessors.go` file.
Specifically:
 - `ReadOnly` type `bool`
 - `Created` type `string`
 - `Modified` type `string`
 - `CreatedBy` type `CreatedBy`

A custom type `CreatedBy` has the following fields:
 - `Disabled` type `bool`
 - `Handle` type `string`
 - `Name` type `string`
 - `IsAdmin` type `bool`
 - `Role` type `string`
 - `AccessRole` type string
 - `Verified` type `bool`
 - `Email` type `string`
 - `Icon` type `string`

PR also includes a unittest.

### DIFF
--- a/dashboards_test.go
+++ b/dashboards_test.go
@@ -2,6 +2,9 @@ package datadog
 
 import (
 	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


dashboards_test.go
```diff
@@ -75,3 +78,109 @@ func (suite *YAxisTestSuite) TestYAxisJSONUnmarshalAutoMinMax() {
...
+	}
+
+	expectedReadOnly := false
+	if readOnly, ok := d1.GetReadOnlyOk(); !ok || readOnly != expectedReadOnly {
+		t.Fatalf("expect read_only exist and be set to %v. Got exists %v, value %v", expectedReadOnly, ok, readOnly)
+	}
+
+	expectedModified := "2018-09-11T06:38:09.000000+00:00"
+	if modified, ok := d1.GetModifiedOk(); !ok || modified != expectedModified {
+		t.Fatalf("expect modified field to exist and be set to %s. Got exists %v, value %s", expectedModified, ok, modified)
```
ojongerius: I see this with the other checks too; if (in this case) `Modified` was *not* set, `GetModifiedOk` will return an empty string and boolean false. In that case this check will fail with (notice value is empty): 

> expect modified field to exist and be set to 2018-09-11T06:38:09.000000+00:00. Got exists false, value

tests/fixtures/dashboards_response.json
```diff
@@ -0,0 +1,44 @@
+{
```
ojongerius: 🥇

ojongerius:
Thanks for the contribution @mnaboka!

dashboards_test.go
mnaboka: fixed!

PR #187, opened 2018-10-25
Add support for units and includeZero in Yaxis

nyanshak:
* Adds IncludeZero to support whether to always include zero or not.
Default is to always include zero.
* Adds IncludeUnits (units in Datadog json) to determine whether to show
the unit label to the left of the yaxis or not.

### DIFF
--- a/dashboards.go
+++ b/dashboards.go
@@ -52,11 +52,13 @@ type GraphEvent struct {
 }
 
 type Yaxis struct {
-	Min     *float64 `json:"min,omitempty"`
-	AutoMin bool     `json:"-"`
-	Max     *float64 `json:"max,omitempty"`
-	AutoMax bool     `json:"-"`
-	Scale   *string  `json:"scale,omitempty"`
+	Min          *float64 `json:"min,omitempty"`
+	AutoMin      bool     `json:"-"`
+	Max          *float64 `json:"max,omitempty"`
+	AutoMax      bool     `json:"-"`
+	Scale        *string  `json:"scale,omitempty"`
+	IncludeZero  *bool    `json:"includeZero,omitempty"`
+	IncludeUnits *bool    `json:"units,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


nyanshak:
Questions for other reviewers:

* I've made Datadog's "units" field into IncludeUnits to hopefully make the intent more clear. Should I change it to Units to more closely match?
* I tested IncludeZero and IncludeUnits with string values as well "true" and "false". Datadog's API accepts and returns them, but the UI does NOT handle them correctly. I believe the most correct thing to do here is keep the type as *bool and fail if a string is received / not support string types. Does that make sense?

gphat:
I came here to make a PR like this only to see that @nyanshak had already done it. I'm not involved in this project but would love to use this PR and am :+1:  to the PR itself. :)

yfronto:
LGTM. The failing tests look like failures for some Go versions due to golint. Probably worth checking if Travis CI needs some other args for those versions in another PR.

> I've made Datadog's "units" field into IncludeUnits to hopefully make the intent more clear. Should I change it to Units to more closely match?

I think `IncludeUnits` is clearer. The struct already differs exactly from the UI (due to the the Auto stuff), so it doesn't bother me.

> I tested IncludeZero and IncludeUnits with string values as well "true" and "false". Datadog's API accepts and returns them, but the UI does NOT handle them correctly. I believe the most correct thing to do here is keep the type as *bool and fail if a string is received / not support string types. Does that make sense?

The type is fine, IMO, unless consumers can get into an ugly state that's hard to debug (like the auto issue). I'd rather the UI behave poorly than have something downstream like Terraform breaks. Also, did you file a bug upstream about it?

nyanshak:
Consumers could get into a weird state if someone else were to edit this incorrectly in the json editor or the API. Unlike the other times we've seen similar problems (like "5" vs 5 in max/min), using string types here does not actually work in the UI.

I just raised this through Datadog support, so... fingers crossed :) Thanks for reviewing, @yfronto

PR #190, opened 2018-10-31
Bugfix fix screenboard

kterada0509:
Issue https://github.com/terraform-providers/terraform-provider-datadog/issues/105

- some screenboard fields are invalid from api response
    - `Width`
        - `string` -> `int`
    - `Height`
        - `string` -> `int`

- Screen Widget json map is invalid
    - `x` and `y` are reversed


kterada0509:
@zorkian 
Thank you for merged.
I have few questions. 
- Is next release already scheduled?
- Are these changes included to it?

PR #193, opened 2018-11-13
Move back Precision to string type

masci:
#179 Changed the type of `Precision` fields for various struct from `*string` to `*json.Number` but I can't make it work due to the fact that `"*"` is a valid value, meaning "full precision". Please notice acceptance tests don't expose the issue since `Precision` field for `Widget` (the one used in the tests) is still a string: https://github.com/zorkian/go-datadog-api/blob/ae3affd1a9e7b1798f38ad1257b5916b7e6ede00/screen_widgets.go#L109

Not sure what was the rationale around #179 since it mentions the API might return `int` for `precision` sometimes but if that's the case we should fix the issue on the Datadog side, normalizing all the things to strings.

This is blocking an important fix for https://github.com/terraform-providers/terraform-provider-datadog/issues/117

### DIFF
--- a/dashboards.go
+++ b/dashboards.go
@@ -123,10 +123,10 @@ type GraphDefinition struct {
 	Yaxis Yaxis `json:"yaxis,omitempty"`
 
 	// For query value type graphs
-	Autoscale  *bool        `json:"autoscale,omitempty"`
-	TextAlign  *string      `json:"text_align,omitempty"`
-	Precision  *json.Number `json:"precision,omitempty"`
-	CustomUnit *string      `json:"custom_unit,omitempty"`
+	Autoscale  *bool   `json:"autoscale,omitempty"`
+	TextAlign  *string `json:"text_align,omitempty"`
+	Precision  *string `json:"precision,omitempty"`
+	CustomUnit *string `json:"custom_unit,omitempty"`
 
 	// For hostmaps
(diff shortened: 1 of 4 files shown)


nyanshak:
I made a small playground link to demonstrate why it should be *json.Number rather than string: https://play.golang.org/p/6TAkg8WrJsg.

Namely, Datadog can return `"precision": 1` (int) rather than `"precision": "1"` (string), which will cause errors with *string, but not *json.Number.

For that reason, I'm rejecting this change.

masci:
@nyanshak try to add `Precision: datadog.JsonNumber(json.Number("*")),` to any `TileDef` in `TestWidgets` to expose the problem. Currently the integration tests don't exercise changes from #179.

Can you please keep this open for discussion, or should I move this to an issue? The current state of the client is half broken (in any case, https://github.com/zorkian/go-datadog-api/blob/ae3affd1a9e7b1798f38ad1257b5916b7e6ede00/screen_widgets.go#L109 is still a `*string`).

nyanshak:
I think you're right that `Precision` field in `Widget` being a different type than in `TileDef` isn't quite right. I have raised #194 to address it. Could you provide feedback there letting me know if that is sufficient?

masci:
Superseded by https://github.com/zorkian/go-datadog-api/pull/194

PR #194, opened 2018-11-13
Make Precision field in Widget *json.Number

nyanshak:
* Make Precision type for Widget match the type for TileDef.

PR #195, opened 2018-11-14
[WIP]Feature/fix width/height data type

kterada0509:
Fixes https://github.com/terraform-providers/terraform-provider-datadog/pull/118

- Datadog `width` param takes int & string like `100%`
- Datadog `height` param also can be int or string

### DIFF
--- a/integration/screenboards_test.go
+++ b/integration/screenboards_test.go
@@ -93,8 +93,8 @@ func TestScreenboardGet(t *testing.T) {
 func getTestScreenboard() *datadog.Screenboard {
 	return &datadog.Screenboard{
 		Title:   datadog.String("___Test-Board___"),
-		Height:  datadog.Int(600),
-		Width:   datadog.Int(800),
+		Height:  datadog.Height("600"),
+		Width:   datadog.Width("100%"),
 		Widgets: []datadog.Widget{},
 	}
 }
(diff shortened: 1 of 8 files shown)


integration/screenboards_test.go
```diff
@@ -93,8 +93,8 @@ func TestScreenboardGet(t *testing.T) {
 func getTestScreenboard() *datadog.Screenboard {
 	return &datadog.Screenboard{
 		Title:   datadog.String("___Test-Board___"),
-		Height:  datadog.Int(600),
-		Width:   datadog.Int(800),
+		Height:  datadog.JsonNumber("600"),
+		Width:   datadog.JsonNumber("800"),
```
masci: Similar to #194, this code will fail downstream during marshalling, right before sending the payload to the client. Try setting `Width:   datadog.JsonNumber("100%"),` here to expose the issue.

kterada0509: Thanks for review.
Fixed it.

screenboards.go
```diff
@@ -9,16 +9,61 @@
 package datadog
 
 import (
+	"encoding/json"
 	"fmt"
 )
 
+// WidthS ...
+type WidthS string
```
masci: Why not using `PrecisionT` from https://github.com/zorkian/go-datadog-api/pull/194?

kterada0509: #194 isn't merged yet.
And using `PrecisionT` as `width` or `height` several difficulty for user to consider its meaning.

If used same struct in whole this package, i think shoud be better assign general name.

masci: #194 was merged, you could change the name of the custom type in this PR

kterada0509: i split PR at #198 .
After merge, i rebase this pr.

PR #196, opened 2018-11-16
Fix #192 screenboard width and height should support string

### LINKED ISSUE #192: Screenboard width and height should stay as type string
Hi all,

#190 changed the type for `Screenboard.Height` and `Screenboard.Width` from `*string` to `*int` (see https://github.com/zorkian/go-datadog-api/pull/190/files#diff-7d71d2abb56836dce85c7c7b114c80dcL20) but I believe they should stay as strings, at least `width` since it can be something like `"100%"`.

masterzen:
Datadog API for screenboards width and height can return either numbers, number as strings (like “1024”) or even strings (like “100%”).

The recent fix #190 changed those fields to be only ints, breaking the compatibility with some datadog screen boards (notably the ones returning "100%").

As suggested in #192 this change fixes the issue by making the fields become `json.Number` which should cover all possible returned values.

### DIFF
--- a/datadog-accessors.go
+++ b/datadog-accessors.go
@@ -6369,18 +6369,18 @@ func (r *Rule) SetTimeframe(v string) {
 }
 
 // GetHeight returns the Height field if non-nil, zero value otherwise.
-func (s *Screenboard) GetHeight() int {
+func (s *Screenboard) GetHeight() json.Number {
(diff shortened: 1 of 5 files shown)


masterzen:
Damned, I hadn't seen @kterada0509 had pushed PR #195 for #192 issue.
/me closing this one.

PR #198, opened 2018-11-28
Fix custom type name for reuse

kterada0509:
# Abstract
Fix Custom data type `PrecisionT` to `StrintD`. (comes from: `String & Int data type for datadog`)
Custom data type `PrecisionT` can reuse for other params.  (like width/heght in #195 ).
But reuse in other param, the data name would confuse users.

### DIFF
--- a/common.go
+++ b/common.go
@@ -1,0 +1,29 @@
+package datadog
+
+import "encoding/json"
+
+// StrintD ...
+type StrintD string
+
+// UnmarshalJSON is a Custom Unmarshal for StrintD. The Datadog API can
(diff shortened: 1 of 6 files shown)


common.go
```diff
@@ -0,0 +1,29 @@
...
+// "*" (string).
+func (s *StrintD) UnmarshalJSON(data []byte) error {
+	var err error
+	var strintNumber json.Number
+	if err = json.Unmarshal(data, &strintNumber); err == nil {
+		*s = StrintD(strintNumber)
+		return nil
+	}
+
+	var strintStr StrintD
```
mrwonko: this has to be a string or we end up in an infinite loop.

PR #202, opened 2019-01-03
Fix screen widget FillMin and FillMax type matching

dabcoder:
`fillMax` and `fillMix` can be decimals, so importing a screenboard that includes those with `terraform import datadog_screenboard.<screenboard_name> <screenboard_id>` would throw an error like:

```
cannot unmarshal number into Go struct field TileDefStyle.fillMax of type string 
```

This PR attempts to fix it, based on https://github.com/zorkian/go-datadog-api/blob/master/dashboards.go#L112-L113. 
~Attempts to fix the type for `PaletteFlip` too.~

### DIFF
--- a/integration/screen_widgets_test.go
+++ b/integration/screen_widgets_test.go
@@ -312,8 +312,8 @@ func TestWidgets(t *testing.T) {
 				Style: &datadog.TileDefStyle{
 					Palette:     datadog.String("hostmap_blues"),
 					PaletteFlip: datadog.String("true"),
-					FillMin:     datadog.String("20"),
-					FillMax:     datadog.String("300"),
+					FillMin:     datadog.JsonNumber("20"),
+					FillMax:     datadog.JsonNumber("300"),
 				},
 			},
 		},
--- a/screen_widgets.go
+++ b/screen_widgets.go
@@ -96,10 +96,10 @@ type TileDefRequestStyle struct {
 }
 
 type TileDefStyle struct {
-	Palette     *string `json:"palette,omitempty"`
-	PaletteFlip *string `json:"paletteFlip,omitempty"`
-	FillMin     *string `json:"fillMin,omitempty"`
-	FillMax     *string `json:"fillMax,omitempty"`
+	Palette     *string      `json:"palette,omitempty"`
+	PaletteFlip *string      `json:"paletteFlip,omitempty"`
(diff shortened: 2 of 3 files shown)


dabcoder:
@nyanshak / @ojongerius Mind reviewing this when you have a chance?

screen_widgets.go
```diff
@@ -96,10 +96,10 @@ type TileDefRequestStyle struct {
 }
 
 type TileDefStyle struct {
-	Palette     *string `json:"palette,omitempty"`
-	PaletteFlip *string `json:"paletteFlip,omitempty"`
```
nyanshak: This should remain `*string` because Datadog's API will accept strings or boolean values. If a user (through the json editor in the UI or via the API), sends `"paletteFlip": "true"`, Datadog will accept and return that value in future calls as a string type, and the UI handles that correctly.

nyanshak:
Happy to accept this if we leave paletteFlip as a `*string` type, or if you have an idea for better handling of string-type paletteFlips.

screen_widgets.go
dabcoder: @nyanshak thanks for the prompt review. Updated the PR and squashed the commits too. Thoughts?

nyanshak: Can you update the integration tests to use the new types for `FillMin` and `FillMax`? Specifically, changing L315-316 in `integration/screen_widgets_test.go` to use `datadog.JsonNumber` rather than `datadog.String`?

dabcoder: Done, thanks

integration/screen_widgets_test.go
```diff
@@ -4,7 +4,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/zorkian/go-datadog-api"
```
nyanshak: Don't remove this line, as it's needed to pull in types:
```
[go-datadog-api] make testacc                                           11:51:09  ☁  dabcoder-master ☂ ✭
go test integration/* -v  -timeout 90m
# command-line-arguments [command-line-arguments.test]
integration/screen_widgets_test.go:10:15: undefined: datadog
FAIL	command-line-arguments [build failed]
make: *** [Makefile:17: testacc] Error 2
```

dabcoder: Not sure what happened there, didn't mean to remove it, but good catch, re-added it.

PR #203, opened 2019-01-11
Adding unaggregated parameter to events.GetEvents

EricByers:
This PR adds support for `unaggregated` parameter for query events.

See: https://docs.datadoghq.com/api/?lang=python#query-the-event-stream

This change is not backwards compatible as it changes the function signature.  This function could potentially be rewritten to a variatiac to keep it backwards compatible if needed.

### DIFF
--- a/events.go
+++ b/events.go
@@ -62,11 +62,12 @@ func (client *Client) GetEvent(id int) (*Event, error) {
 
 // QueryEvents returns a slice of events from the query stream.
 func (client *Client) GetEvents(start, end int,
-	priority, sources, tags string) ([]Event, error) {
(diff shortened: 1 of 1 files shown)


bkabrda:
@EricByers hey :wave: thanks for sending the PR. I'd much rather have a new method, e.g. `GetEventsWithAggregation(start, end int, priority, sources, tags string, unaggregated bool)` and keep the old function intact. I do understand that the variadic argument helps here, but I don't think that this solution scales well with potential new arguments.

The old `GetEvents` function could then just be a proxy to the new one, calling it with `unaggregated=false`. WDYT?

PR #206, opened 2019-02-09
Hyphen should not be used for go filename

dharada1:
This is not big problem but there is no reason not to fix it.


dharada1:
i found it came from gen-accessors, so it is better not to fix it.

Issue #209, opened 2019-02-10
Integration Test Fails : TestMonitorCreateAndDelete IncludeTags


dharada1:
ran on my local, with my datadog api & app key

also master https://travis-ci.org/zorkian/go-datadog-api/builds/490935059 fails

```
=== RUN   TestMonitorCreateAndDelete
--- FAIL: TestMonitorCreateAndDelete (1.26s)
	monitors_test.go:21: 
			Error Trace:	monitors_test.go:21
			Error:      	Not equal: 
			            	expected: &datadog.Monitor{Creator:(*datadog.Creator)(0xc420246720), Id:(*int)(0xc42024a828), Type:(*string)(0xc420218450), Query:(*string)(0xc420218430), Name:(*string)(0xc420218440), Message:(*string)(0xc420218420), OverallState:(*string)(0xc420218460), OverallStateModified:(*string)(nil), Tags:[]string{}, Options:(*datadog.Options)(0xc420160180), State:datadog.State{Groups:map[string]datadog.GroupData(nil)}}
			            	actual  : &datadog.Monitor{Creator:(*datadog.Creator)(0xc420246680), Id:(*int)(0xc42024a718), Type:(*string)(0xc4202185f0), Query:(*string)(0xc420218580), Name:(*string)(0xc4202185a0), Message:(*string)(0xc420218590), OverallState:(*string)(0xc4202185e0), OverallStateModified:(*string)(nil), Tags:[]string{}, Options:(*datadog.Options)(0xc420160400), State:datadog.State{Groups:map[string]datadog.GroupData(nil)}}
			            	
			            	Diff:
			            	--- Expected
			            	+++ Actual
			            	@@ -29,3 +29,3 @@
			            	   ThresholdWindows: (*datadog.ThresholdWindows)(<nil>),
			            	-  IncludeTags: (*bool)(<nil>),
			            	+  IncludeTags: (*bool)(true),
			            	   RequireFullWindow: (*bool)(true),
			Test:       	TestMonitorCreateAndDelete
	monitors_test.go:27: 
			Error Trace:	monitors_test.go:27
			Error:      	Not equal: 
[...]

PR #210, opened 2019-02-10
[Integration Test] Fix fixture : add IncludeTags to test Monitor

dharada1:
https://github.com/zorkian/go-datadog-api/issues/209

go 1.9 test

```
=== RUN   TestMonitorCreateAndDelete
--- PASS: TestMonitorCreateAndDelete (1.44s)
```

"actual" (real datadog API response) returns bool `include_tags`, so "expected" should be fixed.

`include_tags` is `true` when tags exits with the monitor, and `false` when no tags with the monitor.


dharada1:
also it would fix `TestIntegrationGCPCreateAndDelete` failure in go 1.11.x

ojongerius:
Thanks @dharada1 !

Issue #221, opened 2019-03-08
`PUT https://api.datadoghq.com/api/v1/user/<USER_ID>` Seems to be more paranoid now.


brendanjerwin:
We've started getting `400 Bad Request: {"errors": ["Additional properties are not allowed (u'handle' was unexpected)"]}` from the Terraform Datadog provider.

I suspect that the server has recently gotten more paranoid and is now erroring. 
I assume this line: https://github.com/zorkian/go-datadog-api/blob/master/users.go#L88 results in a JSON payload that includes `handle`.

I'm willing to work on a solution. Any guidance? (My first inclination is a struct specifically for the update user action and a factory to create one from the common user struct.)

bkabrda:
Fixed in #222. Closing this issue now.

PR #222, opened 2019-03-08
Remove `handle` from user update

brendanjerwin:
See: #221 

Introduce a struct for the user update request that doesn't include `handle`.

### DIFF
--- a/datadog-accessors.go
+++ b/datadog-accessors.go
@@ -13282,6 +13282,223 @@ func (r *reqSingleServiceLevelObjective) SetData(v ServiceLevelObjective) {
 	r.Data = &v
 }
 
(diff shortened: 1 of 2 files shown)


bkabrda:
LGTM :+1:. @brendanjerwin are you ok with me squashing the commits into a single one while merging?

bkabrda:
This will also need rebasing now, so please do that. Thanks!

dabcoder:
@brendanjerwin Just catching up on that one, can you do the above? Thanks in advance!

brendanjerwin:
There we go. BTW, Datadog loosened the API constraints that originally prompted me to make this change. However, its probably better to not send the useless field anyway...

PR #231, opened 2019-04-08
[dashboard] Add support for the new Dashboard API

MLaureB:
This PR adds support for the [new Dashboard API](https://docs.datadoghq.com/api/?lang=python#dashboards) in the library.

### DIFF
--- a/board_widgets.go
+++ b/board_widgets.go
@@ -1,3 +1,11 @@
+/*
+ * Datadog API for Go
+ *
+ * Please see the included LICENSE file for licensing information.
+ *
+ * Copyright 2019 by authors and contributors.
+ */
+
 package datadog
 
 import (
--- a/board_widgets.go
+++ b/board_widgets.go
@@ -1,0 +1,712 @@
+package datadog
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+const (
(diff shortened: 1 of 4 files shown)


bkabrda:
:+1: LGTM. We'll have to figure out how to properly obsolete and remove things connected to the old APIs, but that's not urgent and it's better done separately anyway.

boards.go
```diff
@@ -0,0 +1,59 @@
+/*
+ * Datadog API for Go
+ *
+ * Please see the included LICENSE file for licensing information.
+ *
+ * Copyright 2013 by authors and contributors.
```
ojongerius: Could do with an update to 2019

board_widgets.go
```diff
@@ -0,0 +1,712 @@
+package datadog
```
ojongerius: I know we've been a little inconsistent with it, but could you please add license information like you did for `board.go`?

boards.go
bkabrda: Updated, resolving

board_widgets.go
bkabrda: Added, resolving.

bkabrda:
@ojongerius all your comments have been addressed, so I'm going to hope that you won't mind me merging before you get to re-review.

Thanks @MLaureB!

PR #232, opened 2019-04-15
Add missing status now returned by the API

tanguylebarzic:
The API changed to always return the status, so the integration test needs to be updated.


gabsn:
Looks good, only affecting the tests.

bkabrda:
LGTM, merging. Thanks for the fix!

PR #234, opened 2019-04-16
Make arguments to APIKey-related methods consistent

bkabrda:
... by accepting values as arguments everywhere.


bkabrda:
Thanks for the review @therve. Merging.

PR #235, opened 2019-04-16
Fix invocation of go vet in Makefile

therve:
go tool vet is not supported since go 1.12, we should use go vet instead. This
fixes the flags used, and all call it for the base package.


bkabrda:
Hmm, travis tests fail with `vet: flag "-structtag" not defined` - could you please investigate?

bkabrda:
LGTM, merging. Thanks!

PR #238, opened 2019-05-13
Make it possible to mute/unmute specific monitor scopes

bkabrda:
This implements the functionality to mute/unmute specific monitor scopes while keeping backwards compatibility. Necessary to fix https://github.com/terraform-providers/terraform-provider-datadog/issues/181


bkabrda:

PR #241, opened 2019-05-16
Add possibility to get/set metadata of graph requests on screenboards and timeboards

bkabrda:
Necessary to implement https://github.com/terraform-providers/terraform-provider-datadog/issues/61

### DIFF
--- a/integration/screen_widgets_test.go
+++ b/integration/screen_widgets_test.go
@@ -45,6 +45,11 @@ func TestWidgets(t *testing.T) {
 						Type:    datadog.String("dashed"),
 						Width:   datadog.String("thin"),
 					},
+					Metadata: map[string]datadog.TileDefMetadata{
+						"avg:system.cpu.user{*}": {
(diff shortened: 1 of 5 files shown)


integration/screen_widgets_test.go
```diff
@@ -45,6 +45,11 @@ func TestWidgets(t *testing.T) {
 						Type:    datadog.String("dashed"),
 						Width:   datadog.String("thin"),
 					},
+					Metadata: map[string]datadog.TileDefMetadata{
```
nmuesch: Is this just testing that a screenboard widget object can be created without error?

bkabrda: If you look at the loop at the end of this test method, it does this:

* Creates an empty screeboard
* Updates it with a definition of an individual widget (=> so creates the widget)
* Reads the whole screenboard back and verifies that it got back exactly what it posted

So this tests both creating and reading back the values.

bkabrda:

PR #243, opened 2019-05-29
Add monitor_id in Synthetics test config

gabsn:
The Synthetics API now returns the monitor ID along with the test config.
This especially allows to use a Synthetics test in a composite monitor.


bkabrda:
LGTM, merging. Thanks!

PR #246, opened 2019-06-10
Fix integration tests for downtimes and synthetics tests after recent changes

bkabrda:


bkabrda:
Thanks for the review. Merging.

PR #247, opened 2019-06-14
Add support for new endpoints that manipulate individual PD service objects

bkabrda:


bkabrda:

PR #248, opened 2019-06-14
As documentation states, PUT has to be used to create the PD integration

bkabrda:
Using POST (the previous state) makes the operation fail sometimes with 405 error, see e.g. https://github.com/terraform-providers/terraform-provider-datadog/issues/160

See https://docs.datadoghq.com/api/?lang=bash#create-a-pagerduty-integration for documentation of this endpoint - PUT is really the suggested verb to use to create the integration.


bkabrda:

PR #252, opened 2019-07-23
[slos] fix API payloads

platinummonkey:
We changed the name of this the threshold `slo` property to `target` as it's more meaningful.  The hardest part of SW - naming.

Also updated the content types of create and added request validation as well.


bkabrda:
The recent changes have removed the `createSLOThreshold` and `createUpdateServiceLevelObjective`, but (some of) the accessors for these remained in the autogenerated accessors file for these, so the compilation fails. Could you please remove these so I can test and merge? Thanks!

platinummonkey:
yup! I'm waiting for us to merge && deploy the final API changes before continuing this work 👌

bkabrda:
@platinummonkey ok, feel free to ping me whenever this is ready to be re-reviewed.

platinummonkey:
@bkabrda good to go 👍 

```
--- PASS: TestServiceLevelObjectivesCreateGetUpdateAndDelete (3.84s)
PASS
```

integration/service_level_objectives_test.go
```diff
@@ -34,32 +35,38 @@ func TestServiceLevelObjectivesCreateGetUpdateAndDelete(t *testing.T) {
 	assert.Equal(t, expected.Description, actual.Description)
 	assert.True(t, expected.Thresholds.Equal(actual.Thresholds))
 
+	time.Sleep(time.Second)
```
bkabrda: Why the sleep? Is it expected that the SLO object is not available through API immediately? Or is it available but incomplete?

platinummonkey: I found that in a few rare cases, the read replicas were behind just enough to make this fail without some artificial delay.

bkabrda: Ok, thanks for the explanation.

bkabrda:
LGTM now, merging. Thanks!

Issue #253, opened 2019-07-31
/v1/dashboard to get all dashboards is missing


Morsicus:
Hey, 

I think this route https://docs.datadoghq.com/api/?lang=python#get-all-dashboards is not implemented in the library, right?

May i help? Or maybe i'm missing something in the documentation :/. 

In any case, please let me know.

bkabrda:
Hey!

Yeah, it's not. TBH I'm not sure why that's the case, but I'd be happy to review a PR if you want to submit support for this. You'd need to add it to https://github.com/zorkian/go-datadog-api/blob/master/boards.go, it should be pretty straightforward.

Thanks for reaching out!

PR #255, opened 2019-08-01
feat: add get all dashboards support (New dashboards API)

Morsicus:
## Features

* Implement `get all dashboards` for `/v1/dashboard` (new dashboards API) -> close https://github.com/zorkian/go-datadog-api/issues/253

### DIFF
--- a/boards.go
+++ b/boards.go
@@ -29,6 +29,24 @@ type Board struct {
 	ModifiedAt        *string            `json:"modified_at,omitempty"`
 }
 
+// BoardLite represents a simplify dashboard (without widgets, notify list, ...)
+// It's used when we load all boards.
+type BoardLite struct {
+	Title        *string `json:"title,omitempty"`
+	Description  *string `json:"description,omitempty"`
+	LayoutType   *string `json:"layout_type,omitempty"`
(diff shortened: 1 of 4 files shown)


boards.go
```diff
@@ -29,6 +29,24 @@ type Board struct {
 	ModifiedAt        *string            `json:"modified_at,omitempty"`
 }
 
+// BoardLite represents a simplify dashboard (without widgets, notify list, ...)
+// It's used when we load all boards.
+type BoardLite struct {
+	Title        *string `json:"title"`
```
bkabrda: Most often, we prefer putting `omitempty` everywhere to stay consistent, I'd suggest you do that.

boards.go
```diff
@@ -29,6 +29,24 @@ type Board struct {
...
+	Description  *string `json:"description,omitempty"`
+	LayoutType   *string `json:layout_type`
```
bkabrda: You forgot quotes around `layout_type` (plus also make it `omitempty`, please).

boards.go
Morsicus: amended

boards.go
Morsicus: amended

PR #256, opened 2019-08-08
[dashboards] Update interval type from int to string in screenboards from legacy api

jbenais:
In the legacy api, the interval type in the compute field of the log query is a string, not an integer.


bkabrda:
LGMT, merging. Thanks!

PR #257, opened 2019-08-08
[dashboard widgets] Add "multi_compute" field to dashboard logs queries

mhintz:
The current "compute" field is a single object, but we need multiple "compute" aggregations for the columns of the table widget. To solve this problem without doing a painful migration of the type of the "compute" field, we decided to add a field called `multi_compute` which contains an array of "compute" objects. This validation is already in place in the dashboards backend, and this PR adds it to the Go API.


bkabrda:
LGTM, merging. Thanks for the PR!

PR #259, opened 2019-08-09
Add IP ranges

dabcoder:
Could be used in the terraform provider as there is an open feature request for that: 
https://github.com/terraform-providers/terraform-provider-datadog/issues/216. 

WIP, so the approach may not be the best.

### DIFF
--- a/request.go
+++ b/request.go
@@ -34,7 +34,7 @@ type Response struct {
 func (client *Client) uriForAPI(api string) (string, error) {
 	var err error
 	// If api is a URI such as /v1/hosts/, /v2/dashboards... add credentials and return a properly formatted URL
-	if !strings.HasPrefix(api, "https://") || !strings.HasPrefix(api, "http://") {
+	if !(strings.HasPrefix(api, "https://") || strings.HasPrefix(api, "http://")) {
 		apiBase, err := url.Parse(client.baseUrl + "/api" + api)
 		if err != nil {
 			return "", err
--- a/request.go
+++ b/request.go
@@ -33,8 +33,8 @@ type Response struct {
 // and it will give the proper request URI to be posted to.
 func (client *Client) uriForAPI(api string) (string, error) {
 	var err error
-	// If api is an endpoint starting with v1, we construct the apiBase URL using the baseUrl to which we append "/api" and the actual value of api
-	if strings.Contains(api, "v1") {
+	// If api is a URI such as /v1/hosts/, /v2/dashboards... add credentials and return a properly formatted URL
+	if !strings.HasPrefix(api, "https://") || !strings.HasPrefix(api, "http://") {
 		apiBase, err := url.Parse(client.baseUrl + "/api" + api)
 		if err != nil {
 			return "", err
--- a/request.go
+++ b/request.go
@@ -28,8 +28,9 @@ type Response struct {
 	Error  string `json:"error"`
 }
 
-// uriForAPI is to be called with something like "/v1/events" and it will give
-// the proper request URI to be posted to.
+// uriForAPI is to be called with either an API resource like "/v1/events"
+// or a full URL like the IP Ranges one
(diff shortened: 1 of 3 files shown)


bkabrda:
Hey David, thanks for submitting this PR. The `GetIPRanges` function should look more like any other function for getting all entities, e.g. have a look at https://github.com/zorkian/go-datadog-api/blob/master/api_keys.go#L74 - Most of the boilerplate code you've written is actually covered by the `client.doJsonRequest`, so I'd suggest using that. Does that make sense?

request.go
```diff
@@ -31,17 +31,43 @@ type Response struct {
 // uriForAPI is to be called with something like "/v1/events" and it will give
 // the proper request URI to be posted to.
 func (client *Client) uriForAPI(api string) (string, error) {
-	apiBase, err := url.Parse(client.baseUrl + "/api" + api)
+	var err error
+	// If api is an endpoint starting with v1, we construct the apiBase URL using the baseUrl to which we append "/api" and the actual value of api
+	if strings.Contains(api, "v1") {
```
bkabrda: TBH I'm a little unsure about this condition. We could theoretically be passing in full URLs that we expect to just return and which might still contain `v1`. I'd much rather that this was `if !(strings.HasPrefix("https://") or strings.HasPrefix("http://"))`

request.go
```diff
@@ -31,17 +31,43 @@ type Response struct {
 // uriForAPI is to be called with something like "/v1/events" and it will give
 // the proper request URI to be posted to.
```
bkabrda: Could you please amend the docstring of this function to cover the change done below?

request.go
dabcoder: Right, I also started in that direction but both URL types will start with `https`:
- `client.baseUrl + "/api" + api` will be `https://api.datadoghq.com/api/v1....`
- IP ranges URLs: `https://ip-ranges.datadoghq.com/` or `https://ip-ranges.datadoghq.eu/`
So not sure if we should compare based on `https` or `http`?

nmuesch: Hey, sorry I'm a bit late here. 
Hmm, this is a good point, because we also aren't sending the API/APP keys in the query params based on this `if` condition. Due to that and the fact that we have some v2 endpoints already (see the dash list v2 endpoint), we can't use this condition, otherwise those requests will be sent without auth and fail. 

Let me know if I'm reading any of this incorrectly 😄 At this point in the code, `api` is either a URI from the caller, like `/v2/dashboards/...` or `/v1/hosts/...` or it's the whole URL for IP ranges `https://ip-ranges.datadoghq.com/`. 
(the `https:api.datadog....` is only added later) So I think the proposed change Slavek made would work here. And then we only add authentication (api/app keys) to paths that are URIs, and not if a full URL is passed into this function.

request.go
```diff
@@ -28,20 +28,47 @@ type Response struct {
...
-// uriForAPI is to be called with something like "/v1/events" and it will give
-// the proper request URI to be posted to.
+// uriForAPI is to be called with either an API resource like "/v1/events"
+// or a full URL like the IP Ranges one
+// and it will give the proper request URI to be posted to.
 func (client *Client) uriForAPI(api string) (string, error) {
-	apiBase, err := url.Parse(client.baseUrl + "/api" + api)
+	var err error
+	// If api is a URI such as /v1/hosts/, /v2/dashboards... add credentials and return a properly formatted URL
+	if !strings.HasPrefix(api, "https://") || !strings.HasPrefix(api, "http://") {
```
bkabrda: So I didn't actually try this out, but I don't think this condition is right. If the `api` value is e.g. `https://something`, the second part would still evaluate as `true` and so the `if` clause would actually be entered. I think what you need to do is `!(strings.HasPrefix(api, "https://") || strings.HasPrefix(api, "http://"))`

PR #260, opened 2019-08-13
Submit custom_fg_color instead of color

nmuesch:
The conditional formats section of screenboards should submit a field called `custom_fg_color` instead of `color`. Docs - https://docs.datadoghq.com/graphing/graphing_json/widget_json/#conditional-format-schema

Should still be backwards compatible since I left the struct alone and only changed what its un-marshalled as.


bkabrda:
LGTM, merging. Thanks for the PR!

PR #262, opened 2019-08-19
Add v2 of the dashboard_list_item API

nmuesch:
Adds the v2 version of the dashboard list item API - https://docs.datadoghq.com/api/?lang=python#get-items-of-a-dashboard-list

Makes new methods for each CRUD operation (same names as the V1 version but appended with `V2`)

Only the individual items API has a V2, so this doesn't duplicate the V1 dashboard_list API. Mostly the same, but the Item struct `ID` field is now all capitalized, and returns a string instead of an integer.

### DIFF
--- a/dashboard_list_items_v2.go
+++ b/dashboard_list_items_v2.go
@@ -12,63 +12,63 @@ import (
 	"fmt"
 )
 
-// DashboardListV2Item represents a single dashboard in a dashboard list.
-type DashboardListV2Item struct {
+// DashboardListItemV2 represents a single dashboard in a dashboard list.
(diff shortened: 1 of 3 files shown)


nmuesch:
@jbenais Thanks for the review. Added a set of tests (its pretty much identical to the tests for v1, but updated to use the new structs/methods) Let me know what you think. 

I also don't have a strong preference but found myself referring to these as `DashboardListV2Items` vs having the `V2` after the word items. Let me know if you think one is preferential to the other. (The API methods are consistent)

bkabrda:
LGTM, merging 👍 Thanks for doing this!

PR #215, opened 2019-02-11
Support for Host totals

dharada1:
https://github.com/zorkian/go-datadog-api/issues/214

support `GET /v1/hosts/totals` .

PR #267, opened 2019-08-23
Add hosts total

dabcoder:
Follow up of #215.


hosts.go
```diff
@@ -1,9 +1,9 @@
 package datadog
 
 type HostActionResp struct {
```
bkabrda: While I do understand that you wanted to make this change to keep things consistent, I would rather not do it. It is backwards incompatible and while it's unfortunate that the code was written this way, let's keep the existing structure as it is to not break any depending code.

dabcoder:
Thanks for the contribution too @dharada1

PR #199, opened 2018-11-28
Add full param options for monitors

Seanstoppable:
Add a more generic function that will let people mix and match
Have existing functions wrap this better function
Resolves the rest of state support in #164

PR #268, opened 2019-08-26
extend monitor api

dbenque:
Extend monitor API to have filtering by tag and group state.


Seanstoppable:
I have a similar PR that adds all options at #199, which (biased opinion) is a little bit more flexible and means we have to make fewer breaking API changes in the future if more options are added.

dbenque:
Yes I agree, your PR is more flexible. I would be happy to discard this one of yours is merged.

dabcoder:
We can close this since we merged https://github.com/zorkian/go-datadog-api/pull/199 then.

PR #270, opened 2019-09-11
Add support for Logs config API

tt810:
Support for Logs public config API `pipeline` and `pipelinelist`. Pipelinelist has **read** and **update** functions.  When a 'pipeline' is created, it's automatically included in the pipelinelist, thus there is no **creation** or **deletion** for pipelinelist.


logs_processor.go
```diff
@@ -0,0 +1,239 @@
...
+	ServiceRemapperType     = "service-remapper"
+	StatusRemapperType      = "status-remapper"
+	TraceIdRemapperType     = "trace-id-remapper"
+	UrlParserType           = "url-parser"
+	UserAgentParserType     = "user-agent-parser"
+)
+
+// LogsProcessor struct represents the processor object from Config API.
+type LogsProcessor struct {
+	Name       *string     `json:"name"`
```
ajacquemot: why do we need points to primitive types ?

logs_processor.go
```diff
@@ -0,0 +1,239 @@
...
+// UserAgentParser represents the user agent parser from config API.
+type UserAgentParser struct {
+	Sources   []string `json:"sources"`
+	Target    *string  `json:"target"`
+	IsEncoded *bool    `json:"is_encoded"`
+}
+
+// convert converts the first argument of type interface{} to a map of string and interface{}
+// and sign the result to the second argument.
+func convert(definition interface{}, processor map[string]interface{}) error {
```
ajacquemot: This method is semi generic, semi specific, e.g. we have generic types with specific argument names, can we make it fully generic or fully specific please ? Also why do we need to marshal unmarshal and not just cast ?

logs_pipeline_list.go
```diff
@@ -0,0 +1,29 @@
+package datadog
+
+const (
+	logsPipelineListPath = "/v1/logs/config/pipeline-order"
+)
```
gaetan-deputier: Structurally it's good, but I would emphasise on the sorted aspect of the pipeline ids in the method names

logs_pipeline.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+)
+
+const (
+	logsPipelinesPath = "/v1/logs/config/pipelines"
+)
+
+// LogsPipeline struct to represent the json object received from Logs Public Config API.
+type LogsPipeline struct {
+	Id         *string              `json:"id,omitempty"`
```
gaetan-deputier: Why would we have omit empty on Id & Type?

logs_pipeline.go
```diff
@@ -0,0 +1,65 @@
...
+		return nil, err
+	}
+	return createdPipeline, nil
+}
+
+// UpdateLogsPipeline updates the pipeline object of a given pipeline id.
+func (client *Client) UpdateLogsPipeline(id string, pipeline *LogsPipeline) (*LogsPipeline, error) {
+	var updatedPipeline = &LogsPipeline{}
+	if err := client.doJsonRequest("PUT", fmt.Sprintf(logsPipelinesPath+"/%s", id), pipeline, updatedPipeline); err != nil {
+		return nil, err
```
gaetan-deputier: nit: Sprintf("%s/%s", logsPipelinesPath, id)

logs_processor.go
```diff
@@ -0,0 +1,239 @@
...
+
+// GrokRule represents the rules for grok parser from config API.
+type GrokRule struct {
+	SupportRules *string `json:"support_rules"`
+	MatchRules   *string `json:"match_rules"`
+}
+
+// NestedPipeline represents the pipeline as processor from config API.
+type NestedPipeline struct {
+	Filter     *FilterConfiguration `json:"filter"`
```
gaetan-deputier: More a product question: We don't have is enabled for a nested pipeline?

logs_processor.go
gaetan-deputier: I tend to agree that the 3 steps (json marshal/unmarshal and all marshal) seems heavy. Isn't there another way to manage it?

logs_processor.go
tt810: I see a couple of things from making the fields all int (as they are pointers):
1. We don't need to copy over the values (e.g. string length can be arbitrary.)
2. The accessors are auto generated from a script (check-code-generation-ran.sh), it could be simpler to expect fields have predictable type.

logs_pipeline.go
gaetan-deputier: Thanks!

datadog-accessors.go
```diff
@@ -1254,6 +1254,254 @@ func (a *APPKey) SetOwner(v string) {
...
+// GetExpression returns the Expression field if non-nil, zero value otherwise.
+func (a *ArithmeticProcessor) GetExpression() string {
+	if a == nil || a.Expression == nil {
+		return ""
+	}
+	return *a.Expression
+}
+
+// GetExpressionOk returns a tuple with the Expression field if it's non-nil, zero value otherwise
+// and a boolean to check if the value has been set.
```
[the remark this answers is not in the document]
bkabrda: These are autogenerated. They may not be strictly necessary, but they're inline with what we do for all the other struct members, so let's keep them for consistency.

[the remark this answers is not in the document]
gaetan-deputier: Fair enough, thanks

PR #272, opened 2019-09-16
Fix integration test for client base url

bkabrda:
Fixes regression introduced by https://github.com/zorkian/go-datadog-api/pull/261


bkabrda:

PR #274, opened 2019-09-30
Update AWS integration endpoints

mzneos:
Add a few missing AWS integration endpoints

### DIFF
--- a/datadog-accessors.go
+++ b/datadog-accessors.go
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
 	i.RoleName = &v
 }
 
+// GetLambdaARN returns the LambdaARN field if non-nil, zero value otherwise.
+func (i *IntegrationAWSLambdaARN) GetLambdaARN() string {
+	if i == nil || i.LambdaARN == nil {
+		return ""
+	}
+	return *i.LambdaARN
+}
+
+// GetLambdaARNOk returns a tuple with the LambdaARN field if it's non-nil, zero value otherwise
+// and a boolean to check if the value has been set.
+func (i *IntegrationAWSLambdaARN) GetLambdaARNOk() (string, bool) {
+	if i == nil || i.LambdaARN == nil {
+		return "", false
+	}
+	return *i.LambdaARN, true
+}
+
+// HasLambdaARN returns a boolean if a field has been set.
+func (i *IntegrationAWSLambdaARN) HasLambdaARN() bool {
+	if i != nil && i.LambdaARN != nil {
+		return true
+	}
+
+	return false
+}
+
+// SetLambdaARN allocates a new i.LambdaARN and returns the pointer to it.
+func (i *IntegrationAWSLambdaARN) SetLambdaARN(v string) {
+	i.LambdaARN = &v
+}
+
+// GetAccountID returns the AccountID field if non-nil, zero value otherwise.
+func (i *IntegrationAWSLambdaARNRequest) GetAccountID() string {
+	if i == nil || i.AccountID == nil {
+		return ""
+	}
+	return *i.AccountID
+}
+
+// GetAccountIDOk returns a tuple with the AccountID field if it's non-nil, zero value otherwise
+// and a boolean to check if the value has been set.
+func (i *IntegrationAWSLambdaARNRequest) GetAccountIDOk() (string, bool) {
+	if i == nil || i.AccountID == nil {
+		return "", false
+	}
+	return *i.AccountID, true
+}
+
+// HasAccountID returns a boolean if a field has been set.
+func (i *IntegrationAWSLambdaARNRequest) HasAccountID() bool {
+	if i != nil && i.AccountID != nil {
+		return true
+	}
+
+	return false
+}
+
+// SetAccountID allocates a new i.AccountID and returns the pointer to it.
+func (i *IntegrationAWSLambdaARNRequest) SetAccountID(v string) {
+	i.AccountID = &v
+}
+
+// GetLambdaARN returns the LambdaARN field if non-nil, zero value otherwise.
+func (i *IntegrationAWSLambdaARNRequest) GetLambdaARN() string {
+	if i == nil || i.LambdaARN == nil {
+		return ""
+	}
+	return *i.LambdaARN
+}
+
+// GetLambdaARNOk returns a tuple with the LambdaARN field if it's non-nil, zero value otherwise
+// and a boolean to check if the value has been set.
+func (i *IntegrationAWSLambdaARNRequest) GetLambdaARNOk() (string, bool) {
+	if i == nil || i.LambdaARN == nil {
+		return "", false
+	}
+	return *i.LambdaARN, true
+}
+
+// HasLambdaARN returns a boolean if a field has been set.
+func (i *IntegrationAWSLambdaARNRequest) HasLambdaARN() bool {
+	if i != nil && i.LambdaARN != nil {
+		return true
+	}
+
+	return false
+}
+
+// SetLambdaARN allocates a new i.LambdaARN and returns the pointer to it.
+func (i *IntegrationAWSLambdaARNRequest) SetLambdaARN(v string) {
+	i.LambdaARN = &v
+}
+
+// GetAccountID returns the AccountID field if non-nil, zero value otherwise.
+func (i *IntegrationAWSLogCollection) GetAccountID() string {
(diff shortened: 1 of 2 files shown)


datadog-accessors.go
```diff
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
...
+// HasLambdaARN returns a boolean if a field has been set.
+func (i *IntegrationAWSLambdaARN) HasLambdaARN() bool {
+	if i != nil && i.LambdaARN != nil {
+		return true
+	}
+
+	return false
+}
+
+// SetLambdaARN allocates a new i.LambdaARN and returns the pointer to it.
```
zippolyte: Doesn't allocate anything nor return anything.
```suggestion
// SetLambdaARN set LambdaARN to point to the given string.
```

datadog-accessors.go
```diff
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
...
+// HasAccountID returns a boolean if a field has been set.
+func (i *IntegrationAWSLambdaARNRequest) HasAccountID() bool {
+	if i != nil && i.AccountID != nil {
+		return true
+	}
+
+	return false
+}
+
+// SetAccountID allocates a new i.AccountID and returns the pointer to it.
```
zippolyte: Doesn't allocate anything nor return anything.
```suggestion
// SetAccountID set AccountID to point to the given string.
```

datadog-accessors.go
```diff
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
...
+// HasLambdaARN returns a boolean if a field has been set.
+func (i *IntegrationAWSLambdaARNRequest) HasLambdaARN() bool {
+	if i != nil && i.LambdaARN != nil {
+		return true
+	}
+
+	return false
+}
+
+// SetLambdaARN allocates a new i.LambdaARN and returns the pointer to it.
```
zippolyte: Doesn't allocate anything nor return anything.
```suggestion
// SetLambdaARN set LambdaARN to point to the given string.
```

datadog-accessors.go
```diff
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
...
+// HasAccountID returns a boolean if a field has been set.
+func (i *IntegrationAWSLogCollection) HasAccountID() bool {
+	if i != nil && i.AccountID != nil {
+		return true
+	}
+
+	return false
+}
+
+// SetAccountID allocates a new i.AccountID and returns the pointer to it.
```
zippolyte: Doesn't allocate anything nor return anything.
```suggestion
// SetAccountID set AccountID to point to the given string.
```

datadog-accessors.go
```diff
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
...
+// HasAccountID returns a boolean if a field has been set.
+func (i *IntegrationAWSServicesLogCollection) HasAccountID() bool {
+	if i != nil && i.AccountID != nil {
+		return true
+	}
+
+	return false
+}
+
+// SetAccountID allocates a new i.AccountID and returns the pointer to it.
```
zippolyte: Doesn't allocate anything nor return anything.
```suggestion
// SetAccountID set AccountID to point to the given string.
```

datadog-accessors.go
```diff
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
 	i.RoleName = &v
 }
 
+// GetLambdaARN returns the LambdaARN field if non-nil, zero value otherwise.
+func (i *IntegrationAWSLambdaARN) GetLambdaARN() string {
```
gzussa: Do we need this one since we have the one below?

datadog-accessors.go
```diff
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
...
+	return false
+}
+
+// SetLambdaARN allocates a new i.LambdaARN and returns the pointer to it.
+func (i *IntegrationAWSLambdaARN) SetLambdaARN(v string) {
+	i.LambdaARN = &v
+}
+
+// GetAccountID returns the AccountID field if non-nil, zero value otherwise.
+func (i *IntegrationAWSLambdaARNRequest) GetAccountID() string {
```
gzussa: Do we need this one since we have the one below?

datadog-accessors.go
```diff
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
...
+	return false
+}
+
+// SetAccountID allocates a new i.AccountID and returns the pointer to it.
+func (i *IntegrationAWSLambdaARNRequest) SetAccountID(v string) {
+	i.AccountID = &v
+}
+
+// GetLambdaARN returns the LambdaARN field if non-nil, zero value otherwise.
+func (i *IntegrationAWSLambdaARNRequest) GetLambdaARN() string {
```
gzussa: Do we need this one since we have the one below?

datadog-accessors.go
```diff
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
...
+	return false
+}
+
+// SetLambdaARN allocates a new i.LambdaARN and returns the pointer to it.
+func (i *IntegrationAWSLambdaARNRequest) SetLambdaARN(v string) {
+	i.LambdaARN = &v
+}
+
+// GetAccountID returns the AccountID field if non-nil, zero value otherwise.
+func (i *IntegrationAWSLogCollection) GetAccountID() string {
```
gzussa: Do we need this one since we have the one below?

datadog-accessors.go
```diff
@@ -10089,6 +10089,161 @@ func (i *IntegrationAWSAccountDeleteRequest) SetRoleName(v string) {
...
+	return false
+}
+
+// SetAccountID allocates a new i.AccountID and returns the pointer to it.
+func (i *IntegrationAWSLogCollection) SetAccountID(v string) {
+	i.AccountID = &v
+}
+
+// GetAccountID returns the AccountID field if non-nil, zero value otherwise.
+func (i *IntegrationAWSServicesLogCollection) GetAccountID() string {
```
gzussa: Do we need this one since we have the one below?

mzneos:
> A few nits on the docstrings. I realize it's the case for all the other accessors, but let's start to be accurate 🙂

this was generated by the command `make generate` so I will try to see if I can fix it directly there (if this is even possible), so that we don't have to do this again

datadog-accessors.go
mzneos: No, I don't think so. This was generated by the command `make generate` which I ran following an error message that in the test pipeline.

I will investiguate to try to understand why there two similar functions were generated

zippolyte:
Okay since this is all autogenerated, changing this is not for the scope of this PR IMO.
Let's leave it like this for now.

mzneos:
> Okay since this is all autogenerated, changing this is not for the scope of this PR IMO.
> Let's leave it like this for now.

Indeed, you're right. Thank you for the review

datadog-accessors.go
bkabrda: Yeah, perhaps we strictly need all of these, but this is how the code generation works right now and it's not in the scope of this PR to investigate/change that. I'm going to mark all conversations for `datadog-accessors.go` as resolved.

integrations.go
```diff
@@ -197,6 +219,12 @@ func (client *Client) CreateIntegrationAWS(awsAccount *IntegrationAWSAccount) (*
 	return &out, nil
 }
 
+// UpdateIntegrationAWS updates an already existing AWS Account in the AWS Integration
+func (client *Client) UpdateIntegrationAWS(awsAccount *IntegrationAWSAccount) error {
+	additionalParameters := "account_id=" + *awsAccount.AccountID + "&role_name=" + *awsAccount.RoleName
```
bkabrda: I'd much rather see usage of `url.Values` here: https://golang.org/pkg/net/url/#Values which will give us safety guarantee and is the canonical way of constructing query strings in go.

mzneos: That is indeed cleaner that what I did in the first place. It's using it now, thank you for the tip!

PR #275, opened 2019-10-08
Add Azure integration

dabcoder:
See https://github.com/terraform-providers/terraform-provider-datadog/issues/196 for context.


integrations.go
```diff
@@ -274,3 +274,68 @@ func (client *Client) UpdateIntegrationGCP(cir *IntegrationGCPUpdateRequest) err
...
+// IntegrationAzureCreateRequest defines the request payload for creating Datadog-Azure CloudPlatform integration.
+type IntegrationAzureCreateRequest struct {
+	TenantName   *string `json:"tenant_name"`
+	ClientID     *string `json:"client_id"`
+	ClientSecret *string `json:"client_secret"`
+	HostFilters  *string `json:"host_filters,omitempty"`
+}
+
+// IntegrationAzureUpdateHostFiltersRequest defines the request payload for creating Datadog-Azure CloudPlatform integration.
+type IntegrationAzureUpdateHostFiltersRequest struct {
```
bkabrda: The API documentation says that `client_secret` is required for this request as well. I'd prefer making all the structures from this PR one - `IntegrationAzure` - perhaps with the exception of `IntegrationAzureUpdateAccountRequest` which has some unique arguments (`new_client_id` and `new_client_name` - BTW I couldn't find any documentation of these in the API docs, where are they defined?).

integrations.go
```diff
@@ -274,3 +274,68 @@ func (client *Client) UpdateIntegrationGCP(cir *IntegrationGCPUpdateRequest) err
...
+	TenantName  *string `json:"tenant_name"`
+	ClientID    *string `json:"client_id"`
+	HostFilters *string `json:"host_filters"`
+}
+
+// IntegrationAzureUpdateAccountRequest defines the request payload for creating Datadog-Azure CloudPlatform integration.
```
bkabrda: This comment seems to be wrong, this is definitely not a create payload.

integrations.go
dabcoder: Yes, so this is what we discussed with @Ricky-Thomas, no documentation for this endpoint just yet. 
I can however work on the `struct`s, thanks for the review.

dabcoder:
Still WIP as this is pending feedback for the `IntegrationAzureUpdateAccountRequest` / Update Azure account.

integrations.go
Ricky-Thomas: There are docs for that endpoint available now here: https://docs.datadoghq.com/api/?lang=bash#integration-azure

Ricky-Thomas:
The docs are currently up to date with all the Azure functionality that should be added here. Lets make sure all the existing functionality is accounted for here.

gzussa:
We (Datadog) implemented this feature in our Official Go Client https://github.com/DataDog/datadog-api-client-go. Closing this PR. If the open source community want to add this feature into this client. Feel free to reopen this PR.

PR #279, opened 2019-10-26
/validate endpoint does not check app key validity

tonglil:
https://docs.datadoghq.com/api/?lang=bash#authentication

This only checks the API key.
You can supply a bad APP key and it will still be valid if only the API key is valid.


bkabrda:
Good catch, thanks for this. Merging.

PR #287, opened 2019-11-20
[logs_processors] Add support for lookup processors

blemale:
Add support for lookup processors, see https://docs.datadoghq.com/logs/processing/processors/?tab=ui#lookup-processor

### DIFF
--- a/logs_processors.go
+++ b/logs_processors.go
@@ -99,7 +99,7 @@ type LookupProcessor struct {
 	Source        *string  `json:"source"`
 	Target        *string  `json:"target"`
 	LookupTable   []string `json:"lookup_table"`
-	DefaultLookup *string  `json:"default_lookup"`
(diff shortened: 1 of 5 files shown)


blemale:
Hi @bkabrda ! Could you have your 👀 on this PR?
Thanks in advance for your time :)

logs_processors.go
```diff
@@ -93,6 +94,14 @@ type GrokRule struct {
 	MatchRules   *string `json:"match_rules"`
 }
 
+// LookupProcessor represents the lookup processor from config API.
+type LookupProcessor struct {
+	Source        *string  `json:"source"`
```
bkabrda: The fields that are not required should be marked as `omitempty` (IIUC that's `default_lookup`)

blemale: You're right! I have made the change 👌

PR #290, opened 2019-12-13
Add new monitor summary widget parameters

chrismdd:
This PR adds two new, optional parameters for the monitor summary (manage status) widget: `summary_type`, and `show_last_triggered`: https://docs.datadoghq.com/graphing/widgets/monitor_summary/#api.

It also updates the `TestWidgets` integration test to test them.


board_widgets.go
```diff
@@ -304,17 +304,19 @@ type LogStreamDefinition struct {
...
-	Count           *int    `json:"count,omitempty"`
-	Start           *int    `json:"start,omitempty"`
-	DisplayFormat   *string `json:"display_format,omitempty"`
-	ColorPreference *string `json:"color_preference,omitempty"`
-	HideZeroCounts  *bool   `json:"hide_zero_counts,omitempty"`
-	Title           *string `json:"title,omitempty"`
-	TitleSize       *string `json:"title_size,omitempty"`
-	TitleAlign      *string `json:"title_align,omitempty"`
+	Type              *string `json:"type"`
+	SummaryType       *string `json:"summary_type"`
```
bkabrda: Is this a required property or not? If not (which I assume is the case), it should be marked with `omitempty` (which is also done is `screen_widgets.go` in this PR).

chrismdd: Good catch! I pushed a commit to fix it.

PR #291, opened 2019-12-26
Adding logic to extract the Rate Limiting statistics

CharlyF:
In an effort to give maximum visibility over the communication between the client and the backend, we would like to introduce logic to extract the Rate Limiting stats from the header of calls made to rate limited endpoints.
More details in the[ official API doc](https://docs.datadoghq.com/api/?lang=bash#rate-limiting).

It exposes a thread safe public method to get the values. 
Additionally, we update the map associated with the Client in the method closest to the request to ensure that we don't miss calls.

### DIFF
--- a/ratelimit.go
+++ b/ratelimit.go
@@ -48,5 +48,10 @@ func (client *Client) updateRateLimits(resp *http.Response, api string) error {
 func (client *Client) GetRateLimitStats() map[string]RateLimit {
 	client.m.Lock()
 	defer client.m.Unlock()
-	return client.rateLimitingStats
+	// Shallow copy to avoid corrupted data
+	mapCopy := make(map[string]RateLimit, len(client.rateLimitingStats))
+	for k, v := range client.rateLimitingStats {
+		mapCopy[k] = v
+	}
+	return mapCopy
 }
--- a/ratelimit.go
+++ b/ratelimit.go
@@ -30,8 +30,8 @@ func isRateLimited(method string, endpoint string) (limited bool, shortEndpoint 
 }
 
 func (client *Client) updateRateLimits(resp *http.Response, api string) error {
-	if resp.Header == nil {
-		return fmt.Errorf("header missing from the HTTP response.")
+	if resp == nil || resp.Header == nil {
+		return fmt.Errorf("could not parse headers from the HTTP response.")
 	}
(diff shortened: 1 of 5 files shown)


ratelimit_test.go
```diff
@@ -0,0 +1,122 @@
...
+	for i, tt := range tests {
+		t.Run(fmt.Sprintf("#%d %s", i, tt.desc), func(t *testing.T) {
+			err := client.updateRateLimits(tt.resp, tt.api)
+			assert.Equal(t, tt.error, err)
+			assert.Equal(t, tt.header, client.rateLimitingStats[tt.api])
+		})
+	}
+}
+
+func makeHeader(limit, period, reset, remaining string) *http.Response {
```
xornivore: What are your thoughts on using that same `RateLimit` struct as a param to this func - these four params here are strings, easy to get confused which one is which?

ratelimit.go
```diff
@@ -0,0 +1,52 @@
...
+		Remaining: resp.Header.Get("X-RateLimit-Remaining"),
+	}
+	return nil
+}
+
+// GetRateLimitStats is a threadsafe getter to retrieve the rate limiting stats associated with the Client.
+func (client *Client) GetRateLimitStats() map[string]RateLimit {
+	client.m.Lock()
+	defer client.m.Unlock()
+	return client.rateLimitingStats
```
clamoriniere: I think you should do a copy of the `Map` before returning it. Else values can be changed/updated in //

CharlyF:
Thank you for the review @gzussa.

> Do we need to make this a cached/async thingy?

I believe clients should be as lightweight and stateless as possible so I do not believe there should be a caching logic introduced. For the async, it is critical to be able to know when this problem arises, my approach was to 1. Get the most up to date values upon demand, 2. Not contribute to the Rate Limit when requesting the statistics (i.e. we don't have a rate limiting endpoint, so you need to query a rate limited endpoint to get them). Exactly as you said:
> you can get latest values independently from your api calls.

With regards to the following,
> Also, I fear we will forget to add new endpoints to the list of supported endpoints.

The alternative would be to parse the header of every call to look for the `X-RATELIMIT-`, but I think having an explicit list is sufficient for now. I am happy to revisit if you think we need to be more future proof.

PR #296, opened 2020-01-03
Add support for Template Variable Presets for Boards

jackellenberger:
## What

image

"Template Variable Presets" aka "Saved Views" are a quick shortcut that
can be added by hand or via json to dashboards in datadog, and appear in
the top left corner.

They allow you to string together common combinations of template
variable values, so if for example you want a "production us-east-2"
view of your dashboard, you could have a "production us-east-2" preset
that sets your template variables to `environment = production` and
`region = us-east-2`.

This PR adds support for template variable presets to Boards (aka
dashboards, not to be confused with screenboards or timeboards, which
are not supported by the API). This will be a necessary step towards
adding template variable preset support to the [datadog terraform
provider](https://github.com/terraform-providers/terraform-provider-datadog).

Cheers!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ testrace:
 	go test -race $(TEST) $(TESTARGS)
 
 fmt:
-	gofmt -w $(GOFMT_FILES)
+	gofmt -w -s $(GOFMT_FILES)
 
 # vet runs the Go source code static analysis tool `vet` to find
 # any common errors.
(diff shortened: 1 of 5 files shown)


jackellenberger:
Of note: this functionality is sort of wonky on DD's side. As an _update_, `template_variable_presets` works everywhere. As a _create_, `template_variable_presets` is valid for both screenboards and timeboards, but only actually does something for timeboards (`layout_type: ordered`). I don't _think_ that's the desired behavior, so I have a message in to DD to verify.

jackellenberger:
Datadog has fixed this strange behavior so this should be good to go.

PR #300, opened 2020-01-17
Webhooks integration - contributor

dabcoder:
Duplicate of #265 if we do not hear back.


integrations.go
```diff
@@ -158,6 +158,52 @@ func (client *Client) DeleteIntegrationSlack() error {
...
+// IntegrationWebhookRequest defines the structure of the a webhook request
+type IntegrationWebhookRequest struct {
+	Webhooks []Webhook `json:"hooks"`
+}
+
+// Webhook defines the structure of the a webhook
+type Webhook struct {
+	Name             *string `json:"name"`
+	URL              *string `json:"url"`
+	UseCustomPayload *string `json:"use_custom_payload"`
```
bkabrda: Only `Name` and `URL` are marked as required in the API doc [1]. This means that all the other attributes should be also marked with `omitempty` JSON tag.

[1] https://docs.datadoghq.com/api/?lang=bash#create-a-webhooks-integration

PR #301, opened 2020-01-20
Fix downtime integration tests by making downtime start/end dynamic

bkabrda:


bkabrda:
Thanks for the reviews, merging.

PR #304, opened 2020-02-20
New data sources for widgets

roughbits01:
Add security signals and RUM events as data sources for Timeseries, Query Value, Table and Top List dashboard widgets.

PS: [The terraform-provider-datadog PR](https://github.com/terraform-providers/terraform-provider-datadog/pull/416)

### DIFF
--- a/integration/dashboards_test.go
+++ b/integration/dashboards_test.go
@@ -150,6 +150,46 @@ func TestDashboardCreateWithProcessQuery(t *testing.T) {
 	assertDashboardEquals(t, actual, expected)
 }
 
+// Create a dashboard with a rum query graph.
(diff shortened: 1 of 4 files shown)


integration/dashboards_test.go
```diff
@@ -334,6 +334,68 @@ func createGraphWithLogQuery() []datadog.Graph {
 	return graphs
 }
 
+func createGraphWithRumQuery() []datadog.Graph {
```
nmuesch: Thanks for these. These should be called inside a method `Test...` similar to https://github.com/zorkian/go-datadog-api/blob/f06edcd25563ca9a485d70a035c372640815a5c7/integration/dashboards_test.go#L134

Issue #310, opened 2020-03-09
[Feature request] Force deletion of monitors


Towerthousand:
Could we support calling monitor deletion with the `force` flag set?

As per the docs: https://docs.datadoghq.com/api/?lang=bash#delete-a-monitor

This would potentially allow this to be used from the Terraform provider implementation, which we would solve https://github.com/terraform-providers/terraform-provider-datadog/issues/439

bkabrda:
Merged in #311, closing. Thanks!

PR #311, opened 2020-03-09
Add force delete monitor

Towerthousand:
This uses the API flag `force=true` to allow clients to delete monitors even if they're referenced from an SLO.

This implements #310


Towerthousand:
@bkabrda does this sound reasonable? :)

bkabrda:
@Towerthousand yeah, this looks really good, merging this now. Thanks!

PR #312, opened 2020-03-11
Fix go regression in json.Number unmarshalling

bkabrda:
This PR fixes regression in unmarshalling of `json.Number` in Go 1.14 as noted at [1]. In short, the value `"auto"` that used to be unmarshalled properly for `json.Number` is now considered an error.

I went through the other places where we use `json.Number` and it doesn't seem they're affected.

[1] https://golang.org/doc/go1.14#encoding/json


bkabrda:
Thanks for the reviews, merging.

PR #313, opened 2020-03-13
Roles support

### LINKED ISSUE #307: Support for roles
It would be great to be able to manage roles in Terraform. As a first step this SDK would need to be updated to allow the API requests related to roles.

Docs: https://docs.datadoghq.com/api/?lang=bash#roles

bkimbrough88:
Enabling the SDK to perform CRUD operations on Roles as well as manage permissions and users on those roles.

Fixes: #307


sendqueery:
This looks good, but appears to be missing support for scoping unless I've missed something?
It looks like it's missing from the API docs, but is mentioned elsewhere in Datadog's docs: https://docs.datadoghq.com/account_management/faq/managing-global-role-permissions/?tab=datadogussite#granting-permissions-within-limited-scopes

From their example:

```
{
    "data": {
        "type": "permissions",
         "id": <PERMISSION_UUID>,
         "scope": {
             "indexes": [
                 "main",
                 "support"
            0]
        }
    }
}
```

bkimbrough88:
You're right, I've added permission scope to the grant permission requests

pietdaniel:
Some of the APIs implemented are not completely stable yet. My concern is that we may be solidifying prematurely here. Namely, the scope field and the org v1 APIs. Is it possible to postpone the implementation of this functionality? the role and permission APIs can be considered stable.

.dependencies/src/github.com/cenkalti/backoff
```diff
@@ -0,0 +1 @@
+Subproject commit 4b4cebaf850ec58f1bb1fec5bdebdf8501c2bc3f
```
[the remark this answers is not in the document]
bkimbrough88: yeah didn't notice that, removed it, not sure when or why that one got added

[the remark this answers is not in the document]
jirikuncar: Thanks!

.gitignore
```diff
@@ -1,2 +1,239 @@
-*.sublime*
-cmd
+# Created by https://www.gitignore.io/api/go,vim,linux,macos,windows,sublimetext,intellij+iml,visualstudiocode
```
jirikuncar: Do you think we could create a separate PR just for a `.gitignore`?

roles_test.go
```diff
@@ -0,0 +1,665 @@
...
+		} else {
+			_, _ = w.Write(response)
+		}
+
+		expectedType := "roles"
+		if requestBody, err := ioutil.ReadAll(r.Body); err != nil {
+			t.Fatal(err)
+		} else {
+			var roleRequest map[string]*Role
+			println(string(requestBody))
```
jirikuncar: do we still need it?

bkimbrough88: good eye, that should not be there

jirikuncar:
@bkimbrough88 is there a way how we could reduce the diff in `datadog-accessors.go` file (maybe reordering methods) to facilitate the review process?

organizations.go
```diff
@@ -0,0 +1,11 @@
+package datadog
+
+type OrganizationResponse struct {
+	Data *Organization `json:"data,omitempty"`
+}
+
+// TODO: Fill this out with everything you can get back when getting an organization
```
jirikuncar, line 7: ❓

pietdaniel (changes requested):
I worry that it is a bit early for the role permission API to be solidified in the public client. Can we postpone implementing this API while we roadmap the APIs definition internally. Roles Users and Permissions are good to go.

roles.go
```diff
@@ -0,0 +1,198 @@
...
+
+type Sort string
+
+const (
+	SortNameAsc        Sort = "name"
+	SortNameDesc       Sort = "-name"
+	SortModifiedAtAsc  Sort = "modified_at"
+	SortModifiedAtDesc Sort = "-modified_at"
+	SortUserCountAsc   Sort = "user_count"
+	SortUserCountDesc  Sort = "-user_count"
```
jirikuncar, line 16: What about adding `SortRolesBy...`?

organizations.go
bkimbrough88: This was just a comment that there's much more to Organizations than an ID, but that's all I needed for this implementation. I can easily delete the comment or create an issue for adding Organizations support.

bkimbrough88:
> I worry that it is a bit early for the role permission API to be solidified in the public client. Can we postpone implementing this API while we roadmap the APIs definition internally. Roles Users and Permissions are good to go.

I feel like anything posted in https://docs.datadoghq.com/api should be stable enough to write into an SDK. The caveat being that you then have to maintain that feature in the SDK as the API evolves.

bkimbrough88:
Apologies on letting this sit, I'll take care of the PR comments later tonight

PR #319, opened 2020-03-29
Adding unaggregated parameter to events.GetEvents

praveenRamalingam:
Implement @bkabrda suggestion from #203  to add support for querying unaggregated query param


jirikuncar:
@praveenRamalingam the solution looks good. Thank you! Can you also include new test for it?

Issue #322, opened 2020-04-21
Allow insecure certificates is not supported


matt-canty-dragon:
`allow_insecure` is not supported in Synthetics.

Expected as _option_ here:
https://github.com/zorkian/go-datadog-api/blob/master/synthetics.go#L54-L63

Documentation on the topic:
https://docs.datadoghq.com/synthetics/api_tests/?tab=httptest

![image](https://user-images.githubusercontent.com/56073682/79842607-f704ef80-83b0-11ea-805f-19931609cfff.png)

nsuarezcanton:
I've made a quick PR for this issue: https://github.com/zorkian/go-datadog-api/pull/324. Thanks for surfacing this!

PR #289, opened 2019-12-09
Add Retry to Synthetics

s33dunda:
Very simple addition to allow users to add retry logic to their Synthetics Checks.
Being that the api is missing this in their documentation. The payload was gathered via the request  posted while using the UI. Tested and is working with an in house terraform-datadog-provider.

Fixed a couple of typos too.

PR #324, opened 2020-04-28
Add support for Allow Insecure in Synthetics tests

nsuarezcanton:
- A small update to add support for `allow_insecure` within Synthetics tests.
- This aims to address this other issue: https://github.com/zorkian/go-datadog-api/issues/322.
- Seems like no further updates shall be required but let me know if I've missed something. I've used this one as a reference: https://github.com/zorkian/go-datadog-api/pull/289.


nsuarezcanton:
Perfect, thanks for the review! We'll be updating the API docs as well.

PR #325, opened 2020-06-24
Add SearchMonitors to monitors.go

afarbos:
Add Support for Monitor Search, see https://docs.datadoghq.com/api/v1/monitors/#monitors-search

Note: OverallStateModified is not a string when using the monitors/search but an int


afarbos:
stale

